PR #4, opened 2024-01-19
Make Settings page use controlled components

markdavella:
- Created utility for detect when strings contain only numbers
- Made nearly all setting form fields controlled
- experimenting with styling of overall setting pages/tabs

### DIFF
--- a/web2/src/helpers/util.ts
+++ b/web2/src/helpers/util.ts
@@ -50,3 +50,7 @@ export const fromStringResolution = (
   const [h, w] = res.split('x', 1);
   return { widthPx: parseInt(w), heightPx: parseInt(h) };
 };
+
+export const hasOnlyDigits = (value: string) => {
+  return /^-?\d+$/.test(value);
+};
--- a/web2/src/pages/settings/PlexSettingsPage.tsx
+++ b/web2/src/pages/settings/PlexSettingsPage.tsx
@@ -27,6 +27,7 @@ import {
   Typography,
   Input,
   InputAdornment,
+  SelectChangeEvent,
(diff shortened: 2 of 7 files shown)


web2/src/helpers/util.ts
```diff
@@ -50,3 +50,7 @@ export const fromStringResolution = (
   const [h, w] = res.split('x', 1);
   return { widthPx: parseInt(w), heightPx: parseInt(h) };
 };
+
+export const hasOnlyDigits = (value: string) => {
+  return /^-?\d+$/.test(value);
```
chrisbenincasa, line 55: should prob add the global flag to the regex: `/^-?\d+$/g`

web2/src/pages/settings/PlexSettingsPage.tsx
```diff
@@ -119,6 +209,50 @@ export default function PlexSettingsPage() {
...
+
+  const [maxPlayableResolution, setMaxPlayableResolution] = useState<string>(
+    toStringResolution(defaultPlexSettings.maxPlayableResolution),
+  );
+
+  const handleMaxPlayableResolution = (event: SelectChangeEvent<string>) => {
+    setMaxPlayableResolution(event.target.value as string);
+  };
+
+  const [maxDirectStreamBitrate, setMaxDirectStreamBitrate] = useState<string>(
```
chrisbenincasa, line 242: super nit - you shouldnt have to be explicit about the type if it is obvious. vscode/typescript will infer the type as string in this case

chrisbenincasa (approved):

PR #53, opened 2024-02-07
Rewrite to save programs API

chrisbenincasa:
* Supports saving lineups generated by time-slot algo
* "Manually" built lineups are sent without program repeats - saves on
  upload bandwidth to server
* "Time slot" built lineups are regenerated server-side based on the
  schedule so we're not serializing/sending 1000s of programs to the
server at once
* Still some weird logic around updating channel/program associations
  that we should look at
* DB files are now saved in 'minified' JSON format


markdavella (approved):
🚢

PR #72, opened 2024-02-13
Rewrite plex servers API

chrisbenincasa:
* Zod API defintions for all endpoints
* Make delete by ID
* Option at the backend level to not clear out programs (need to expose
  this on API, if we think it is useful)
* Fix on server deletion to not clear program associations if the server
  delete fails
* Use zodios client on frontend for new Plex server endpoint defs
* Other random cleanup as I saw it


markdavella (approved):
You're a mad man!

Issue #79, opened 2024-02-14
Edit Channel page defaulting fields to wrong channels


markdavella:
This happened previously but appears to be happening again. 

<img width="1558" alt="image" src="https://github.com/chrisbenincasa/tunarr/assets/6005331/a075e41e-deef-4173-b71f-6be117133db6">

PR #82, opened 2024-02-14
Fixes issue with loading incorrect channel details on edit page

markdavella:
- Fixes https://github.com/chrisbenincasa/tunarr/issues/79

### DIFF
--- a/web2/src/pages/channels/EditChannelPage.tsx
+++ b/web2/src/pages/channels/EditChannelPage.tsx
@@ -124,9 +124,6 @@ export default function EditChannelPage({ isNew }: Props) {
   });
 
   useEffectOnce(() => {
-    console.log('it ran!');
-    // console.log(channel);
-
     setCurrentChannel(channel);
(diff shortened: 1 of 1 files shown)


web2/src/pages/channels/EditChannelPage.tsx
```diff
@@ -115,7 +114,7 @@ export default function EditChannelPage({ isNew }: Props) {
     // Change this so we only load the form on initial...
     // eslint-disable-next-line @typescript-eslint/require-await
     defaultValues: async () => {
-      const c = originalChannel ?? channel;
+      const c = previousChannel ?? channel;
```
chrisbenincasa, line 117: i have a hunch this works because previousChannel is actually always undefined. Wonder if we can just make this channel

markdavella: lol, that's because you are correct!  I am going to close this and push the fix in a new PR

PR #118, opened 2024-02-21
Create SECURITY.md

Lovebugpaulnice:
### Explanation of the changes, problem that they are intended to fix.

...

### All Submissions:

* [ ] I have read the code of conduct.
* [ ] I am submitting to the correct base branch

### Changes that modify the db structure

* [ ] Backwards compatibility: Users running the new code using an existing .disquetv folder will not lose their channels / settings.
* [ ] I've implemented the necessary db migration steps if any.

### New features

* [ ] I understand that the feature may not be accepted if it doesn't fit the upstream app's planned design direction. But that in this case I am encouraged to share this as an available modification other users can use if they want.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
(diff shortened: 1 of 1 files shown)


chrisbenincasa:
Hey thanks for the contribution. This looks like the default template as the versions don't match. Since we're pre-release I don't think we're quite ready for this anyway, but we will definitely look into adding this in the future!

PR #135, opened 2024-02-28
Implement plex server editing. Fixes #123

### LINKED ISSUE #123: Ability to edit existing Plex servers
In the original version, you can edit Plex server details including name, url, and token. We should add similar functionality (sometimes the derived Plex URL is not the one the user wants to use, e.g. if they have internal DNS in their home setup or somewthing)

chrisbenincasa:


markdavella (approved):
🚢

PR #160, opened 2024-03-08
Improve performance of random slots page

chrisbenincasa:
* Move random slots to their own subcomponent (prevents re-render of whole page when weights change)
* Only update underlying form value on slider commit value (onmouseup)
* Memoize the dropdown portion of each slot row (prevent re-render when sliders change)
* Add CSS animation on non-active sliders, smooth transitions while we're dragging

### DIFF
--- a/web/src/pages/channels/RandomSlotEditorPage.tsx
+++ b/web/src/pages/channels/RandomSlotEditorPage.tsx
@@ -13,7 +13,6 @@ import {
   InputLabel,
   MenuItem,
   Select,
-  SelectChangeEvent,
   Slider,
   Snackbar,
   Stack,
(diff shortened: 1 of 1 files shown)


web/src/pages/channels/RandomSlotEditorPage.tsx
```diff
@@ -639,7 +677,7 @@ export default function RandomSlotEditorPage() {
                 <Select
                   label="Distribution"
                   {...field}
-                  onChange={(e) => handleDistributionChange(e, field.onChange)}
+                  // onChange={(e) => handleDistributionChange(e, field.onChange)}
```
markdavella, line 680: Just a heads up that this is commented out

chrisbenincasa: Ah yep, this was on purpose.

PR #172, opened 2024-03-13
Custom Show Programming Picker - next steps

chrisbenincasa:
Fairly big overhaul of the way the programming picker to start to generalize over different types of media sources. This lays the groundwork for future features, like adding different media server types (or supporting local files without a media server at all)

This includes the beginning implementation of choosing the dizqueTV concept of 'Custom Shows' in the program picker, just like you could do in dizque. There is sitll some work to be done on actually adding the result of the selected programming to the current channel, since that is still Plex specific. That will happen in a follow-up PR.


web/src/components/channel_config/ProgrammingSelector.tsx
```diff
@@ -89,11 +91,7 @@ export default function ProgrammingSelector({ onAddSelectedMedia }: Props) {
   });
 
   useEffect(() => {
-    if (
-      mediaSource === 'custom-shows' &&
-      customShows &&
```
markdavella, line 94: Q. if useCustomShows fails won't customShows be undefined? or because you pass in the initial data it will always exist?

markdavella (approved):
LGTM, just 1 Q non-blocker

PR #191, opened 2024-03-20
Add simple error page - fixes #121

### LINKED ISSUE #121: Add better ErrorBoundary stuff for react-router
Right now if anything goes wrong we just hit this plain white page with 'error' and nothing happens... we should be smarter about handling errors and attempt to retry, etc. And also show something pretty.

chrisbenincasa:
Fixes #121

Adds a simple error page (screenshot below) at the root router level. Employs a mini-hack to make the error page appear in the overall Layout. The downside to this hack, which is probably something we should think about, is that if the error occurs in the layout (Root) component itself, we'll have an issue rendering this Error page. We could consider a barebones Error Root component; perhaps one that makes no network calls or does anything fancy... but we can do that later.

This also includes a mechanism for helping users file bugs for issues they experiencing by auto-generating a URL specific to the experienced error that prefills contextual details (browser, OS, Tunarr versions, etc).

Screenshot from 2024-03-20 10-30-59


web/src/assets/error_this_is_fire.png
markdavella, line 1: nite: this file isn't used anywhere

web/src/assets/error_this_is_fine.png
markdavella, line 1: nit: there is 1px of nose on the left haha

web/src/assets/error.png
markdavella, line 1: nit: this file isn't used

markdavella (approved):
Just a couple non-blocking nits, LGTM!

web/src/assets/error_this_is_fire.png
chrisbenincasa: Thought we might randomly switch between them lmao then I forgot

web/src/assets/error_this_is_fine.png
chrisbenincasa: lol dammit that is so annoying. I'll fix before merging

chrisbenincasa: Fixed

web/src/assets/error.png
chrisbenincasa: Moving this over to design folder - just since it's the original asset and I dont wanna lose it just yet .

PR #195, opened 2024-03-20
Add Inline Modal to Grid View

markdavella:
Added support for an initial version of an Inline Grid Modal

Preview:
image

Future improvements needed:
- Performance improvements, especially when initially opening the modal on a new row.  I believe this is mostly caused by the transition not knowing the height of the modal as its loading in.
- Aesthetics
- Add support for Seasons/Episodes


chrisbenincasa:
Dude this is sick!!

PR #227, opened 2024-04-02
Fixes fade in grid items and some other modal issues

markdavella:
- Fade now correctly works for PlexGridtem component
- Updated PlexGridItem to initiate Fade after the image has fully loaded (so Fade is visible to user)
- Fixes issue where rowSize calculation would be wrong, now rounding the number to make it match Grid calc.
- Update calculation for estimating inline modal height so it doesn't display as the wrong size initially while images load in
- Cleaned up some variable naming in the modal utilities to make it easier to understand


markdavella:
I closed this in favor of the changes I pushed here:
https://github.com/chrisbenincasa/tunarr/pull/233

Issue #93, opened 2024-02-16
Implement remaining programming scheduling features


chrisbenincasa:
<img width="1489" alt="Screenshot 2024-02-16 at 8 26 14 AM" src="https://github.com/chrisbenincasa/tunarr/assets/1640671/ac786d50-48b9-44d0-b88f-e74d29e7e72e">

<img width="817" alt="Screenshot 2024-02-16 at 8 25 49 AM" src="https://github.com/chrisbenincasa/tunarr/assets/1640671/b1689d36-01ba-40a7-8328-062d83747389">

chrisbenincasa:
We have most of these done. From what i can tell, this is what is left:

- [ ] cyclic sort
- [ ] sort TV shows

PR #269, opened 2024-04-11
Add Cyclical Shuffle & Sort by TV Shows

markdavella:
- Fixes https://github.com/chrisbenincasa/tunarr/issues/93
- Light clean up of programming 'add' options

### DIFF
--- a/web/src/hooks/programming_controls/useCyclicShuffle.ts
+++ b/web/src/hooks/programming_controls/useCyclicShuffle.ts
@@ -1,0 +1,30 @@
+import _ from 'lodash-es';
+import useStore from '../../store/index.ts';
+import { materializedProgramListSelector } from '../../store/selectors.ts';
+
+export function useCyclicShuffle() {
+  const programs = useStore(materializedProgramListSelector);
+
+  return function () {
+    console.log(programs);
+
+    function chunkEpisodesByShow(programs) {
+      return _.groupBy(programs, (episode) => episode.showId || 'no_show_id');
+    }
+
+    // Group episodes by show ID
+    const showGroups = chunkEpisodesByShow(programs);
+
+    // Print the grouped episodes
+    console.log(showGroups);
+
+    // let showList = chain(programs)
+    //   .filter(isUIContentProgram)
+    //   .filter((program) => program.subtype === 'episode')
+    //   .value();
+
+    // const finalProgramList = concat(alternatingShows, movieList); // Append movies to the end of the list
+
+    // setCurrentLineup(finalProgramList, true);
+  };
+}
--- a/web/src/hooks/programming_controls/useCyclicShuffle.ts
+++ b/web/src/hooks/programming_controls/useCyclicShuffle.ts
@@ -1,5 +1,5 @@
 import { isContentProgram } from '@tunarr/types';
-import _, { chain } from 'lodash-es';
+import _ from 'lodash-es';
 import { setCurrentLineup } from '../../store/channelEditor/actions.ts';
 import useStore from '../../store/index.ts';
 import { materializedProgramListSelector } from '../../store/selectors.ts';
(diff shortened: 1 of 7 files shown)


web/src/hooks/programming_controls/useEpisodeNumberSort.ts
```diff
@@ -0,0 +1,42 @@
...
+  sortOrder: SortOrder,
+) => {
+  // Extract movies since they are appended at the bottom of the list
+  const movieList = programs.filter(
+    (program) => program.type === 'content' && program.subtype === 'movie',
+  );
+
+  // Sort shows by showId so all unique shows are together
+  // Then sort by season and episode number so they are in the correct order
+  const showList = chain(programs)
```
chrisbenincasa, line 30: Importing the named `chain` from `lodash-es` was the problem I was telling you about the other day. Not sure if it's the same when bundling with vite, chain depends on all of the lodash subfunctions and esbuild was tree-shaking them out when I was using the named chain import. I had to fix this by importing the default `_` import and using chain from there to default "function not found" errors on the final bundle

web/src/hooks/programming_controls/useCyclicShuffle.ts
```diff
@@ -0,0 +1,61 @@
...
+    // randomly grab from each chunk until nothing remains
+
+    // Group shows by showId
+    const sortedPrograms = chain(programs)
+      .filter(isContentProgram)
+      .orderBy(['showId', 'seasonNumber', 'episodeNumber'])
+      .value();
+
+    const groupedContent = _.groupBy(sortedPrograms, (program) => {
+      if (program.type === 'content' && program.subtype === 'episode') {
```
chrisbenincasa, line 23: nit - could add the same for 'track' here and group by album id

web/src/hooks/programming_controls/useEpisodeNumberSort.ts
```diff
@@ -0,0 +1,42 @@
+import { chain, concat } from 'lodash-es';
+import { setCurrentLineup } from '../../store/channelEditor/actions.ts';
+import useStore from '../../store/index.ts';
+import { materializedProgramListSelector } from '../../store/selectors.ts';
+import { UIChannelProgram, isUIContentProgram } from '../../types/index.ts';
+
+export type SortOrder = 'asc' | 'desc';
```
chrisbenincasa, line 7: note to self - we should probably just put this in the common types package and reference it everywhere, i'm sure we have it in a few places already

web/src/hooks/programming_controls/useEpisodeNumberSort.ts
```diff
@@ -0,0 +1,42 @@
...
+  const showList = chain(programs)
+    .filter(isUIContentProgram)
+    .filter((program) => program.subtype === 'episode')
+    .orderBy(
+      ['showId', 'seasonNumber', 'episodeNumber'],
+      [sortOrder, sortOrder, sortOrder],
+    )
+    .value();
+
+  const newProgramSort = concat(showList, movieList); // Append movies to the end of the list
```
chrisbenincasa, line 39: note that as-is, this will remove all 'track's from the program

web/src/hooks/programming_controls/useEpisodeNumberSort.ts
markdavella: I can get this updated in this PR

web/src/hooks/programming_controls/useEpisodeNumberSort.ts
markdavella: Word, just added in something to maintain tracks, will push shortly

web/src/hooks/programming_controls/useEpisodeNumberSort.ts
markdavella: Gotcha, will update to _.chain to be safe

PR #291, opened 2024-04-14
Guide Channel Menu

markdavella:
- Added menu for quick channel navigation to the Guide page
- Updated 'Watch' Icon to be a Play Icon on guide page and on channels page

image

### DIFF
--- a/web/src/pages/guide/GuidePage.tsx
+++ b/web/src/pages/guide/GuidePage.tsx
@@ -181,17 +181,14 @@ export default function GuidePage() {
   >();
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
   const open = Boolean(anchorEl);
+  const darkMode = useStore((state) => state.theme.darkMode);
 
   const handleClick = (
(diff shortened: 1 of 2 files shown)


web/src/pages/guide/GuidePage.tsx
```diff
@@ -415,6 +498,53 @@ export default function GuidePage() {
...
+          disableRipple
+          to={`/channels/${channelMenu.id}/watch`}
+          component={RouterLink}
+        >
+          <WatchIcon />
+          Watch Channel
+        </MenuItem>
+        <MenuItem
+          disableRipple
+          to={`http://localhost:8000/media-player/${channelMenu.number}.m3u`}
```
markdavella, line 538: @chrisbenincasa Do we have a util or function somewhere that can dynamically grab the port that should be used here?

chrisbenincasa: Unfortunately not yet. I was wondering whether we want to expose server path as an option on the frontend but it's sorta a can of worms to do that... let's open an issue for the port specitically

PR #308, opened 2024-04-17
Mark program list as dirty on successfull drag'n'drop. Fixes #307

### LINKED ISSUE #307: Manual DND sorting doesn't make programming 'dirty'
With this change (https://github.com/chrisbenincasa/tunarr/pull/306) we disable the Save button based on whether programming is dirty.  This works great when you use a sort from the drop down, add content, etc.  However, if you manually sort via drag and drop the dirty state never updates.

This issue is more noticeable because of this ux change but the core issue exists outside of that change.  Right now in Main if you manually sorted you would be unable to reset it. 

```
const programsDirty = useStore((s) => s.channelEditor.dirty.programs);
```

chrisbenincasa:


markdavella (approved):
LGTM

PR #317, opened 2024-04-17
docker ffmpeg updates

jasongdove:
This updates the base (edge) docker image to include ffmpeg, and to look more like the NVIDIA one that appears to be working. I ran into entrypoint problems (would drop me in node or ffmpeg) without the same full-stack stuff at the end.

It also updates the NVIDIA docker image to use a base image compiled against CUDA 11 https://github.com/ErsatzTV/ErsatzTV-ffmpeg/blob/ae3e97bc86a5c1eb78dbe401a4c4a04cca4d8938/nvidia.Dockerfile#L1 vs CUDA 12 https://github.com/jrottenberg/ffmpeg/blob/416b0cae4ab3f251385eee9f89c22e66efca6bfd/docker-images/6.1/nvidia2204/Dockerfile#L9 which greatly lowers the NVIDIA driver requirements. Note that this also changes from Ubuntu 22.04 to Ubuntu 20.04.

### DIFF
--- a/build.Dockerfile
+++ b/build.Dockerfile
@@ -55,15 +55,21 @@ RUN pnpm turbo --filter=@tunarr/server bundle
 FROM sources AS build-web
 # Install deps
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
-# Build common modules
+# Build and bundle
 RUN pnpm turbo --filter=@tunarr/web bundle
+
+### Experimental: Build a SEA
+FROM build-server AS build-exec
+COPY --from=build-server /tunarr/server/node_modules /tunarr/server/node_modules
+COPY --from=build-server /tunarr/server/build /tunarr/server/build
+RUN pnpm run --filter=server make-exec
+###
 
 FROM sources as build-full-stack
 # Install deps
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
 # Build common modules
 RUN pnpm turbo bundle
-
 
 ### Begin server run ###
 FROM ffmpeg-base AS server
--- a/build.Dockerfile
+++ b/build.Dockerfile
@@ -1,15 +1,35 @@
-FROM node:20-alpine3.19 AS base
+# Setup a node + ffmpeg base
+FROM jasongdove/ersatztv-ffmpeg:7.0 AS ffmpeg-base
 
-# Update
-RUN apk add --no-cache libc6-compat
-RUN apk update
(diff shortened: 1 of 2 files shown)


build.Dockerfile
```diff
@@ -35,29 +55,34 @@ RUN pnpm turbo --filter=@tunarr/server bundle
 FROM sources AS build-web
 # Install deps
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
-# Build and bundle
+# Build common modules
 RUN pnpm turbo --filter=@tunarr/web bundle
 
-### Experimental: Build a SEA
```
jasongdove, line 41: I admittedly don't understand what this is doing, and just made the file look like the NVIDIA one since things seemed to stop working after switching to a base image with ffmpeg.

nvidia.Dockerfile
```diff
@@ -36,20 +37,11 @@ COPY types ./types
 COPY web ./web
 COPY patches ./patches
 
-FROM ffmpeg-base as dev
```
jasongdove, line 39: I don't think this is necessary, but again, happy to be corrected.

build.Dockerfile
chrisbenincasa: This is a build target for 'bundling' the standalone executable file. I had it available in here so we could reuse the cached layers from previous steps that actually build the app bundle itself. This way we could pretty quickly produce both binaries and docker images

chrisbenincasa: The Nvidia container doesn't need this target since the standalone executable has nothing Nvidia-specific packaged in it. We could theoretically just make another Dockerfile specific to bundling up the standalone executable and have it reuse -- but we'd have to address the layer from build.Dockerfile by it's SHA and it felt a little messy to me.

nvidia.Dockerfile
chrisbenincasa: I'd like to keep this around - it allows me to develop directly in the docker container (that has the pre-built ffmpeg) without having to get it up-and-running on whatever system I am developing on. That said, I found out the version of ffmpeg that you can install with snap has nvenc support, so I have been using that - but I think it is useful to have a dev entrypoint if we ever had to debug issues specific to nvidia (or the container runtime) and our dev systems didn't match what the user was running.

jasongdove: Ah, I see that in the `dev.compose.yml` now :+1:

chrisbenincasa (approved):
Looks great! Thanks for this. Going to merge it and test it out.

PR #346, opened 2024-04-19
Use 'hls/' as hls_base_url

chrisbenincasa:
Changes the session to return /media-player/:channelId/hls as the 'serverPath' used by clients who start their HLS session using the PUT /media-player/:channelId/hls endpoint (in theory we could probably get rid of this completely now and just have the client reissue a GET request to the same endpoint, but this gives us some future flexibility)

Sessions started directly on GET /media-player/:channelId/hls should
work as intended now, too.

image

### DIFF
--- a/server/src/api/hlsApi.ts
+++ b/server/src/api/hlsApi.ts
@@ -173,7 +173,6 @@ export const hlsApi: RouterPluginAsyncCallback = async (fastify) => {
       }
 
       if (req.query.direct) {
-        console.log(session.serverPath);
         return res.redirect(302, `${session.serverPath}?token=${token}`);
       }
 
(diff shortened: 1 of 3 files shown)


jasongdove:
Looks good to me on web and standalone players - the only other change I would make is to set the MIME type on the m3u8 so it works with `ffplay` which refuses to play it currently.

>% ffplay -hide_banner 'http://localhost:8000/media-player/8f16804e-5435-4188-adda-c4c85f4e8400/hls'
Not detecting m3u8/hls with non standard extension and non standard mime type
    Last message repeated 1 times
http://localhost:8000/media-player/8f16804e-5435-4188-adda-c4c85f4e8400/hls: Invalid data found when processing input

It should be `application/vnd.apple.mpegurl` according to https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Delivery

chrisbenincasa:
Ah yep, good call. Added the change.

PR #361, opened 2024-04-23
Add VAAPI/QSV Dockerfile

jasongdove:
Copied `nvidia.Dockerfile` to `vaapi.Dockerfile` and adjusted the base image to enable VAAPI and QSV encoders in docker containers. You just need to pass the appropriate device to docker, e.g. `--device /dev/dri/renderD128:/dev/dri/renderD128`. Many Intel devices can use VAAPI encoders, and Intel devices that use the iHD driver can usually use QSV encoders. AMD devices can use VAAPI encoders. Capabilities can be checked using `vainfo` from inside the container, or on the host.

With that said, I don't believe VAAPI or QSV encoders will work due to the current ffmpeg pipeline in tunarr. NVIDIA happens to work because it's more flexible about automatically uploading frames to hardware for the encoder; VAAPI and QSV are much more finicky. Nonetheless, this might help with testing your ffmpeg pipeline refactoring.

### DIFF
--- a/vaapi.Dockerfile
+++ b/vaapi.Dockerfile
@@ -1,0 +1,94 @@
+# Setup a node + ffmpeg + nvidia base
+FROM jasongdove/ersatztv-ffmpeg:7.0-vaapi AS ffmpeg-base
+ENV NODE_MAJOR=20
+ENV TUNARR_BIND_ADDR=0.0.0.0
+EXPOSE 8000
+
+# Install musl for native node bindings (sqlite)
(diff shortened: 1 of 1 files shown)


chrisbenincasa:
Thanks! Going to merge as-is (with the warning that the current ffmpeg code forked from DTV likely won't work). I will try these out today for QSV on my machine.

I think we should be able to consolidate these Dockerfiles and parameterize on the base image, too. I will work on that after merging here.

PR #382, opened 2024-04-25
Add next/prev buttons to new Channel page. Save button now called 'Finish'

chrisbenincasa:
…nish'

Cleaned up the way this works a bit -- using the custom context and moving the action buttons to the top level page instead of having to duplicate them. 

image


web/src/components/channel_config/ChannelEditActions.tsx
```diff
@@ -5,57 +5,88 @@ import Stack from '@mui/material/Stack';
...
+          type="submit"
+          startIcon={
+            isSubmitting ? (
+              <CircularProgress size="20px" sx={{ mx: 1, color: '#fff' }} />
+            ) : (
+              <Save />
+            )
+          }
+        >
+          Finish
```
markdavella, line 53: I've been tormented by the UX in this section, so hard to make it feel all connected and part of the same flow.  I'm wondering if "finish" should be "Save & Exit" or something like that.

web/src/components/channel_config/ChannelEditActions.tsx
```diff
@@ -5,57 +5,88 @@ import Stack from '@mui/material/Stack';
...
-          >
-            Save
+  const renderNewChannelActions = () => {
+    const prevButton = !isUndefined(channelEditorState.currentTab.prev) ? (
+      <Button onClick={() => onNav(channelEditorState.currentTab.prev!)}>
+        Previous
+      </Button>
+    ) : null;
+    const nextButton = !isUndefined(channelEditorState.currentTab.next) ? (
+      <Button onClick={() => onNav(channelEditorState.currentTab.next!)}>
```
markdavella, line 32: Maybe we add `startIcon` arrows for the next/previous buttons to make it feel more connected?

markdavella (commented):
I love that you made this form feel more connected between the tabs, just added a couple thoughts and ideas.

Should we add an alert above these sections for new Channels that just says something like, "These settings can all be edited later"?  When I create something new in a new platform I am always afraid I won't be able to go back and change some selection I make.

Issue #370, opened 2024-04-25
Add default blank state for TV Guide


markdavella:
Right now TV just shows spinner and eventually "An error occurred: Network Error" when there are 0 channels configured.  We should have a default state when there are no channels.

PR #387, opened 2024-04-26
Add New Component NoChannelsCreated

markdavella:
- New component created to display a default state for TV Guide & Channels page for when no channels exist yet. Directing users to create their first channel
- Fixes: https://github.com/chrisbenincasa/tunarr/issues/370

image

image

### DIFF
--- a/web/src/components/channel_config/NoChannelsCreated.tsx
+++ b/web/src/components/channel_config/NoChannelsCreated.tsx
@@ -1,0 +1,43 @@
+import SettingsRemoteIcon from '@mui/icons-material/SettingsRemote';
+import { Box, Button, Typography } from '@mui/material';
+import { Link as RouterLink } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


web/src/components/channel_config/NoChannelsCreated.tsx
```diff
@@ -0,0 +1,43 @@
+import SettingsRemoteIcon from '@mui/icons-material/SettingsRemote';
+import { Box, Button, Typography } from '@mui/material';
+import { Link as RouterLink } from 'react-router-dom';
+import { useChannels } from '../../hooks/useChannels';
+import PaddedPaper from '../base/PaddedPaper';
+
+export default function NoChannelsCreated() {
```
chrisbenincasa, line 7: Super nit - i think the new consensus is to prefer named exports (no default) because it makes it easier for tooling to identify what you are trying to import

PR #405, opened 2024-04-29
Re-implement XMLTV generation using typesafe @iptv/xmltv library

chrisbenincasa:
- **Checkpoint**
- **Re-implement XMLTV generation using typesafe @iptv/xmltv library**


chrisbenincasa:
Some really basic stats -- I generated a month long XMLTV doc in less than a second here. I didn't measure memory differences between the two implementations. Unfortunately neither seem to support streaming the writes to the output file

PR #482, opened 2024-06-06
Implement 'select all' functionality in program selector

### LINKED ISSUE #481: "Select All" for Programming Channels
Would be a nice quality of life to have a Select All option for media as you are populating a channel. 

I like to make Genre Channels for movies that may have a large quantity of movies in them randomly playing but currently having to select each individual movie is tough. The Filtering works beautifully and is awesome. Just need a way to pulling everything that us Filter once set. 

Let me know if I am missing something as well!

chrisbenincasa:
Closes #481

This adds new functionality to 'select all' of the visible items in the
program selector. The use-case here is to execute a Plex search/filter
and then add all of the results at once

Also includes the following changes:
* Refactor giant plex hooks file into multiple smaller files
* Breakout some common plex search queries into new hooks
* Use FixedList in the selected programming list drawer - I noticed a
  slowdown after implementing select all; this helps.
* Refactor the way the SelectedProgrammingList component works a bit.
  We're no longer mounting this in App with hard-coded paths -- there's
more fixes here, though. Now we have a temporary drawer with various
toggle states, etc etc. This should free up the viewport to view content
* Fixes a bug in Plex search/filtering for TV shows
* Enables the use of path aliases in imports using "@/", since we're
  starting to deal with a lot of files in many nested directories. This
cleans things up a bit


josway:
Awesome...Thanks!

chrisbenincasa:
No problem! Let me know how it goes.

josway:
The Select All is working well but when I go to Save the Channel I made I get an error, below, and the channel won't save. 

Zodios: Invalid Body parameter 'body'

Unsure if they are related or not.

chrisbenincasa:
Probably not related. We’re still hammering out some types. If you’re able
to grab the entire error from the browser console log that would be great.
There may be logs on the server too.

josway:
From Browser. Didn't see anything in the Server Logs. The below error shows up a handful of times in the Browser. 

[Error]  (2)
Error: Zodios: Invalid Body parameter 'body'
Zd — index.mjs:4:1014
(anonymous function) — index.mjs:4:2598
[ { "code": "invalid_type", "expected": "string", "received": "undefined", "path": [ "programs", 404, "originalProgram", "Media", 0, "container" ], "message": "Required" }, { "code": "invalid_type", "expected": "string", "received": "undefined", "path": [ "programs", 404, "originalProgram", "Media", 0, "Part", 0, "container" ], "message": "Required" } ]
	onError (index-b84a391f.js:363:111283)
	(anonymous function) (index-b84a391f.js:192:22856)

chrisbenincasa:
Thanks. This is very helpful!

chrisbenincasa:
Opened #484 to track. Plex's types are pretty complex. We generally just need to be permissive for certain types/fields so this should be a simple fix

PR #522, opened 2024-06-14
Set a dynamic defaultRoute 

markdavella:
- If ShowWelcome is true, default to the Welcome page, otherwise go to Guide
- Update the icon to go to the defaultPath
- Update selected language when nothing is selected so people know why button is disabled

### DIFF
--- a/web/src/router.tsx
+++ b/web/src/router.tsx
@@ -40,8 +40,14 @@ import {
   newFillerListLoader,
 } from './preloaders/fillerListLoader.ts';
 import { queryCache } from './queryClient.ts';
+import useStore from './store/index.ts';
 
 const queryClient = new QueryClient({ queryCache });
+
+const DefaultRoute = () => {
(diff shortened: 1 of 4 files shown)


web/src/router.tsx
```diff
@@ -40,9 +40,15 @@ import {
   newFillerListLoader,
 } from './preloaders/fillerListLoader.ts';
 import { queryCache } from './queryClient.ts';
+import useStore from './store/index.ts';
 
 const queryClient = new QueryClient({ queryCache });
 
+const DefaultRoute = () => {
+  const showWelcome = useStore((state) => state.theme.showWelcome);
```
chrisbenincasa, line 48: FWIW we can read state outside of components like `useStore.getState().theme.showWelcome`

example: https://github.com/chrisbenincasa/tunarr/blob/becdefef4a410b2c844b564aa206a7306418db7d/web/src/preloaders/channelLoaders.ts#L115

Issue #24, opened 2024-01-30
Jellyfin source library support


chrisbenincasa:

binarygeek119:
just to addon to this, emby support should be easy. as a few projects ive seen can just work with little tweaking from emby. not that i need emby support as ive switched to jellyfin. jellyseerr was able to work with emby just fine but did needa little bug fixed that was related to a slightly different url path.

ghzgod:
"Access your channels by adding the spoofed Tunarr HDHomerun tuner to Plex, Jellyfin, or Emby. Or utilize generated M3U files with any 3rd party IPTV player app."

I don't see Jellyfin on the frontend setup page, Is Jellyfin implemented on the backend somewhere?

chrisbenincasa:
Jellyfin is not implemented as a source yet, but you can add your channels (backed by Plex) to Jellyfin as a Live TV source: https://jellyfin.org/docs/general/server/live-tv/setup-guide/

We hope to start implementation of Jellyfin as a programming source soon, though!

ghzgod:
Awesome thank you!

RealBradCool:
how long is this complete

chrisbenincasa:
This is in progress right now. You can follow https://github.com/chrisbenincasa/tunarr/pull/633 for updates on the implementation. There is no ETA right now. So far the following things are implemented:

- Adding jellyfin servers as a source
- UI for browsing your JF libraries 
- Selecting media to add to channels

What's left (might not be a complete list)
- Saving JF media to channels
- Streaming implementation 
- Assorted another cleanup 

I'd say we're still a bit away, but making solid progress

RealBradCool:
um its draft could need fixed and says failed

chrisbenincasa:
Yes I understand it's a draft. It's unfinished work that I push to continuously.

RealBradCool:
hold on maybe i could help

PR #633, opened 2024-07-24
feat!: add support for Jellyfin media

chrisbenincasa:
This commit includes a huge amount of changes, including support for
adding Jellyfin servers as media sources and streaming content from
them.

These are breaking changes and touch almost every corner of the code,
but also pave the way for a lot more flexibility on the backend for
addinng different sources.

The commit also includes performance improvements to the inline modal,
lots of code cleanup, and a few bug fixes I found along the way.

If you are using Tunarr for streaming Plex at version <=0.7.0, it is *highly recommended* to make a backup before upgrading to this version. The DBs are not backwards compatible and it is possible there are breaking bugs here, considering the size of this change. Make a backup so you can revert to 0.7.0 if something goes wrong!

Fixes https://github.com/chrisbenincasa/tunarr/issues/24


RealBradCool:
hold on maybe try working on a sportts channeel maker

RealBradCool:
also why says draft

chrisbenincasa:
Because it is not done or ready.

perkmi:
Leaving this note for anyone else that may find it useful: In Docker (at least for me) the breaking changes for this commit were resolved by deleting "stream-cache.json" and re-adding the media to my channels. I don't have any watermarks or FLEX media on my channels so I can't speak to those. Since the releases page just says "Breaking Changes" I wasn't sure what to expect, but it's relatively minor. Thanks @chrisbenincasa for all your hard work!

chrisbenincasa:
Appreciate the note @perkmi! I hadn't seen anything like that in my testing, especially with the stream-cache, but it's hard to cover all scenarios. Do you happen to have any logs from when you were having issues? It would be awesome to see if I could make the transition smoother.

perkmi:
[tunarr.log](https://github.com/user-attachments/files/16743089/tunarr.log)

Here's the relevant porion of the log.

I didn't do too much troubleshooting so when I saw the error related to "stream-cache.json" I just blew it away and restarted the container. For all I know it could have resolved itself eventually. It didn't prevent tunarr from starting or anything like that.

I'm guessing all the "Was unable to convert lineup item to guide item" messages were related to the channel media. The channels would show the correct "# Programs" on the "Channels" page, but when I clicked on them the "Channels > Programming" were blank. Re-adding the media to the channels sorted that issue out.

chrisbenincasa:
Hmm interesting, thanks for the logs. I wonder if something went wrong with the DB migration that was required for this change. That is really the only breaking change. The "Was unable to convert lineup item to guide item" log can happen for a lot of reasons, but generally it will happen if the guide builder cannot locate a program with a given ID in the database. This would explain why all of the programming was missing in your channels as well. 

I can look deeper tomorrow. I'm not sure where the migration might've went wrong there, since the changes to the programs table are pretty straightforward ...

chrisbenincasa:
I think I have isolated the problem. It seems to be an issue with the DB migration and the foreign key constraints between the programs and channel_programs table. I have a fix in the works. Sorry you had to recreate your channels!

perkmi:
> I think I have isolated the problem. It seems to be an issue with the DB migration and the foreign key constraints between the programs and channel_programs table. I have a fix in the works. Sorry you had to recreate your channels!

Great! and no worries, I spend more time tinkering than watching anyway.

PR #665, opened 2024-08-06
Update setup.md

perkmi:
- Update instructions for Hardware encoding.
- Add volume in docker config to access Tunarr config files.

Just a few lines added to the readme with some info I had to figure out experimentally. No problem if you don't like my wording and want to change it, but it may be helpful for others.


chrisbenincasa:
Thanks!

PR #744, opened 2024-09-03
Add Programming Speed Dial

markdavella:
- Adds Speed Dial to Programming Selector Page
- Fixes selected items count on Programming Selector Page
- Moved tanstack dev tools to bottom left

image


web/src/components/channel_config/SelectedProgrammingActions.tsx
```diff
@@ -140,101 +160,125 @@ export default function SelectedProgrammingActions({
     }
   };
 
+  const addSelectedItems: MouseEventHandler = (e) => {
+    e.preventDefault();
+    e.stopPropagation();
+    setIsLoading(true);
+    sequentialPromises(
+      selectedMedia,
+      forSelectedMediaType<Promise<AddedMedia[]>>({
```
chrisbenincasa, line 169: this looks like it's missing the Jellyfin case -- the existing impl of this has it. maybe we should just factor the existing one out to a common location and reuse it?

Issue #858, opened 2024-10-11
Always fetch complete rows in grid view


markdavella:
We should fetch items that are perfectly divisible by the number of columns, so when they load in its always full rows

Example:
![image](https://github.com/user-attachments/assets/a810e769-5463-41a6-aa6f-3dfcb2ef52e1)

chrisbenincasa:
Reopening just so we don't forget to do Jellyfin too

PR #948, opened 2024-11-08
Fetch Complete Rows in Grid

markdavella:
We now estimate the number of columns based on the screen size and fetch a value divisible by that so that we always have a nice clean grid.

I also added in logic so that if a user resizes their screen and breaks that clean grid, their next fetch of items will grab additional items to accommodate the blank spaces and make the grid perfect again.

Fixes: https://github.com/chrisbenincasa/tunarr/issues/858


chrisbenincasa:
Should we make similar changes to the Jellyfin side of things?

PR #1015, opened 2024-12-11
refactor: implement full IoC pattern

chrisbenincasa:
* Removes as many "new"s as I could find
* Removes default parameters in constructors
* Introduces Provider pattern
* Removes calls to getDatabase() around codebase

All in preparation for proper testing and mock injection


chrisbenincasa:
Pivoted to inversify: https://github.com/chrisbenincasa/tunarr/tree/inversify

PR #1039, opened 2025-01-03
feat(ffmpeg): add audio language preferences

### LINKED ISSUE #1005: Language settings
We should support preferred language settings. This will inform which audio stream (and eventually which subtitle stream) to pick.

AugusDogus:
- Add language preferences to FFmpeg settings to control audio stream selection
- Implement language matching logic in both FFmpeg implementations
- Add UI component for managing language preferences with ISO 639-1 codes
- Update PlexStreamDetails to correctly map language fields
- Add migration to set default English language preference

BREAKING CHANGE: FFmpeg settings schema updated to include languagePreferences

Resolves #1005

image

### DIFF
--- a/server/src/ffmpeg/FfmpegStreamFactory.ts
+++ b/server/src/ffmpeg/FfmpegStreamFactory.ts
@@ -1,13 +1,13 @@
 import { Channel } from '@/db/schema/Channel.ts';
 import { TranscodeConfig } from '@/db/schema/TranscodeConfig.ts';
-import { HttpStreamSource } from '@/stream/types.ts';
+import { AudioStreamDetails, HttpStreamSource } from '@/stream/types.ts';
 import { Maybe, Nullable } from '@/types/util.ts';
 import { isDefined, isLinux, isNonEmptyString } from '@/util/index.ts';
 import { makeLocalUrl } from '@/util/serverUtil.ts';
 import { FfmpegSettings } from '@tunarr/types';
 import dayjs from 'dayjs';
 import { Duration } from 'dayjs/plugin/duration.js';
-import { find, first, isUndefined } from 'lodash-es';
+import { isUndefined } from 'lodash-es';
 import { DeepReadonly } from 'ts-essentials';
 import { FfmpegPlaybackParamsCalculator } from './FfmpegPlaybackParamsCalculator.ts';
 import { FfmpegProcess } from './FfmpegProcess.ts';
@@ -52,7 +52,7 @@ export class FfmpegStreamFactory extends IFFMPEG {
   private pipelineBuilderFactory: PipelineBuilderFactory;
 
   constructor(
-    private ffmpegSettings: DeepReadonly<FfmpegSettings>,
+    private ffmpegSettings: FfmpegSettings,
     private transcodeConfig: TranscodeConfig,
     private channel: Channel,
   ) {
@@ -333,10 +333,10 @@ export class FfmpegStreamFactory extends IFFMPEG {
 
     let audioInput: AudioInputSource;
     if (isDefined(streamDetails.audioDetails)) {
-      const audioStream =
-        find(streamDetails.audioDetails, { selected: true }) ??
-        find(streamDetails.audioDetails, { default: true }) ??
-        first(streamDetails.audioDetails);
+      // Find the best matching audio stream based on language preferences
+      const audioStream = this.findBestAudioStream(
+        streamDetails.audioDetails as AudioStreamDetails[],
+      );
       const audioStreamIndex = isNonEmptyString(audioStream.index)
         ? parseInt(audioStream.index)
         : 1;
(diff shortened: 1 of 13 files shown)


AugusDogus:
I am absolutely not a designer and this was my first time using MUI so feel free to leave feedback regarding the UI.

web/src/components/LanguagePreferencesList.tsx
```diff
@@ -0,0 +1,91 @@
+import languages from '@cospired/i18n-iso-languages';
```
chrisbenincasa, line 1: Sorry that i haven't standardized this in a pre-commit yet (I think it's in a commit somewhere...) but can you format these files with prettier using the configuration at the root of the repo?

server/src/stream/types.ts
```diff
@@ -53,6 +53,8 @@ export type AudioStreamDetails = {
   selected?: boolean;
   title?: string;
   language?: string;
+  languageTag?: string;
```
chrisbenincasa, line 56: language tag is plex terminology right? let's try to keep this agnostic to media source. i'd be OK with just `language2Letter` and `language3Letter`, but open to other ideas, too

server/src/stream/plex/PlexStreamDetails.ts
```diff
@@ -339,7 +339,9 @@ export class PlexStreamDetails {
           // to pick these streams.
           selected: audioStream.selected,
           default: audioStream.default,
-          language: audioStream.languageCode,
+          language: audioStream.language,
+          languageTag: audioStream.languageTag,
```
chrisbenincasa, line 343: I think we'll need a similar change in JellyfinStreamDetails (of course, using the relevant fields from their API)

server/src/ffmpeg/FfmpegStreamFactory.ts
```diff
@@ -333,10 +333,10 @@ export class FfmpegStreamFactory extends IFFMPEG {
 
     let audioInput: AudioInputSource;
     if (isDefined(streamDetails.audioDetails)) {
-      const audioStream =
-        find(streamDetails.audioDetails, { selected: true }) ??
-        find(streamDetails.audioDetails, { default: true }) ??
-        first(streamDetails.audioDetails);
+      // Find the best matching audio stream based on language preferences
+      const audioStream = this.findBestAudioStream(
+        streamDetails.audioDetails as AudioStreamDetails[],
```
chrisbenincasa, line 338: we should avoid casting like this - the audioDetails field is a non-empty array type purposefully. you should be able to take in a `NonEmptyArray<AudioStreamDetails>` (NonEmptyArray is an import from ts-essentials) or just the explicit type in your function and it will behave the same as a regular array.

web/src/components/LanguagePreferencesList.tsx
```diff
@@ -0,0 +1,91 @@
...
+import en from '@cospired/i18n-iso-languages/langs/en.json';
+import { Add as AddIcon, Delete as DeleteIcon } from '@mui/icons-material';
+import { Autocomplete, Box, Button, IconButton, Stack, TextField, Typography } from '@mui/material';
+
+// Initialize the languages database with English names
+languages.registerLocale(en);
+
+export interface LanguagePreference {
```
chrisbenincasa, line 9: [nit] can reuse the shared type here

web/src/components/LanguagePreferencesList.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return (
+        <Box>
+            <Typography variant="h6" sx={{ mt: 2, mb: 1 }}>Audio Language Preferences</Typography>
+
+            <Typography variant="subtitle2" sx={{ mb: 1, fontStyle: 'italic' }}>
+                In order of priority
+            </Typography>
+            <Stack spacing={2}>
+                {preferences.map((pref, index) => (
+                    <Stack key={index} direction="row" spacing={2} alignItems="center">
```
chrisbenincasa, line 52: let's use the language 2-letter code as the key here. Some details on why to not use the 'index' as a key here: https://react.dev/learn/rendering-lists#why-does-react-need-keys

web/src/components/LanguagePreferencesList.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    return (
+        <Box>
+            <Typography variant="h6" sx={{ mt: 2, mb: 1 }}>Audio Language Preferences</Typography>
+
+            <Typography variant="subtitle2" sx={{ mb: 1, fontStyle: 'italic' }}>
+                In order of priority
+            </Typography>
+            <Stack spacing={2}>
+                {preferences.map((pref, index) => (
```
chrisbenincasa, line 51: I think something like this is well-suited for the 'autocomplete with multiple values' pattern: https://react.dev/learn/rendering-lists#why-does-react-need-keys

Happy to merge this as-is though and clean it up later.

AugusDogus:
Thanks for reviewing so promptly! I have a fix for most of these done, but I'm still taking a look at DeepReadonly issue. I believe this was already type-erroring whenever I first pulled the repo without making any changes.

chrisbenincasa:
All commits / PRs are type-checked, so it would be pretty surprising if a fresh pull of dev was erroring (here is the latest run from a few hours ago, ignore the test failure; the type checking passed): https://github.com/chrisbenincasa/tunarr/actions/runs/12652208286

chrisbenincasa:
I'm also a little confused why the default actions didn't run on this pull request...

chrisbenincasa:
Took a closer look at the DeepReadonly issues - it seems that the issue here was that ffmpegSettings only ever defined primitive types and TS treats `readonly string === string`, for instance (makes sense; we're not dealing with references here). So once the language preferences object schema was added to ffmpegSettings, the DeepReadonly check (which we want) kicked in. It's a little inconvenient, because all users of the ffmpegSettings would need to be updated

chrisbenincasa:
Superseded by #1046 - thanks again for you contribution! I've mentioned you in the changelog and release as well!

PR #1040, opened 2025-01-07
feat: improvements to random slot weighting

chrisbenincasa:
Includes the ability to keep weights locked to one another - or by
default, keep them "unlocked" (the classic DTV experience). This is
mainly a UX feature, but serves as the basis for implementing certain
programming tools as random slots (coming in a follow-up). Additionally,
we show actual weights of generated schedules in a tooltip.


cameronurnes:
Thanks!

chrisbenincasa:
@cameronurnes - for sure. Look out for https://github.com/chrisbenincasa/tunarr/pull/1041 which will offer cyclic shuffle as a random slot scheduler preset option.

PR #1140, opened 2025-04-05
feat: add flag for enabling trust proxy

ronhuang:
When running Tunarr behind a reverse proxy server for HTTPS support, users will need to enable "trust proxy" for the generated XMLTV and M3U files to contain the correct protocol (e.g., `http` or `https`) and host. Without "trust proxy" enabled, Tunarr will not honor the `X-Forwarded-Proto` and `X-Forwarded-Host` headers forwarded from the reverse proxy.

Users can enable "trust proxy" either with the `--trustProxy` flag or the `TUNARR_SERVER_TRUST_PROXY` environment variable.

More information on the trust proxy is available at https://github.com/fastify/fastify/blob/main/docs/Reference/Server.md#trustproxy.


chrisbenincasa:
Awesome! Thanks so much for the contribution. I'll take a closer look ASAP

PR #1295, opened 2025-07-31
feat: add genres endpoint and integrate genre selection in the ui for…

devbjackson:
… jellyfin programming

Add ability to filter programming items on a Jellyfin library by genre on the Add Media screen
Snag_6c9b257
Snag_6c9bfc5

### DIFF
--- a/server/src/external/jellyfin/JellyfinApiClient.ts
+++ b/server/src/external/jellyfin/JellyfinApiClient.ts
@@ -359,6 +359,26 @@ export class JellyfinApiClient extends BaseApiClient {
     return `${this.options.url}/web/#/details?id=${id}`;
   }
 
+  async getGenres(
+    parentId?: string,
+    includeItemTypes?: string
+  ): Promise<QueryResult<string>> {
+    const genresResult = await this.doGet<string>({
+      url: `/Genres`,
+      params: {
+        ...(parentId ? { parentId } : {}),
+        includeItemTypes,
+        recursive: 'true',
+      },
+    });
+
+    if (isError(genresResult)) {
+      return this.makeErrorResult('generic_request_error');
+    }
+
+    return this.makeSuccessResult(genresResult);
+  }
+
   async recordPlaybackStart(itemId: string, deviceId: string) {
     return this.doPost({
       url: '/Sessions/Playing',
--- a/server/src/external/jellyfin/JellyfinApiClient.ts
+++ b/server/src/external/jellyfin/JellyfinApiClient.ts
@@ -366,13 +366,14 @@ export class JellyfinApiClient extends BaseApiClient {
     const genresResult = await this.doGet<string>({
       url: `/Genres`,
       params: {
-        ...(parentId ? { parentId } : {}),
+        parentId,
+        userId: this.options.userId,
         includeItemTypes,
         recursive: 'true',
       },
     });
 
-    if (isError(genresResult)) {
+    if (genresResult instanceof Error) {
       return this.makeErrorResult('generic_request_error');
     }
 
--- a/types/src/jellyfin/index.ts
+++ b/types/src/jellyfin/index.ts
@@ -859,6 +859,16 @@ export type JellyfinLibraryItemsResponse = z.infer<
   typeof JellyfinLibraryItemsResponse
 >;
 
+export const JellyfinGenresResponse = z.object({
+  Items: z.array(JellyfinItem),
+  TotalRecordCount: z.number(),
+  StartIndex: z.number().nullable().optional(),
+});
+
+export type JellyfinGenresResponse = z.infer<
+  typeof JellyfinGenresResponse
+>;
+
 const JellyfinSessionInfo = z
   .object({
     // PlayState: PlayerStateInfo.nullable().optional(),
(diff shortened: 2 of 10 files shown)


server/src/external/jellyfin/JellyfinApiClient.ts
```diff
@@ -359,6 +359,26 @@ export class JellyfinApiClient extends BaseApiClient {
...
   }
 
+  async getGenres(
+    parentId?: string,
+    includeItemTypes?: string
+  ): Promise<QueryResult<string>> {
+    const genresResult = await this.doGet<string>({
+      url: `/Genres`,
+      params: {
+        ...(parentId ? { parentId } : {}),
```
chrisbenincasa, line 369: I think we may also wanna add `userid: this.options.userId` here

chrisbenincasa: Also should be able to simplify to just `parentId,` here instead of using the spread

types/src/jellyfin/index.ts
```diff
@@ -859,6 +859,16 @@ export type JellyfinLibraryItemsResponse = z.infer<
   typeof JellyfinLibraryItemsResponse
 >;
 
+export const JellyfinGenresResponse = z.object({
```
chrisbenincasa, line 862: I know the naming isn't ideal, but can we just reuse the `JellyfinLibraryItemsResponse` schema above? Everything in the JF API is a "JellyfinItem"

web/src/components/channel_config/ProgrammingSelector.tsx
```diff
@@ -442,6 +451,54 @@ export const ProgrammingSelector = ({
     }
   };
 
+  /**
+   * Helper function to extract genre name from different API response formats
+   * Jellyfin API can return genres as objects with 'Name' property or as strings
+   */
+  const extractGenreName = (genre: unknown): string => {
```
chrisbenincasa, line 458: We should be able to get a proper type here - wouldn't this a be `JellyfinItem`?

web/src/components/channel_config/ProgrammingSelector.tsx
```diff
@@ -442,6 +451,54 @@ export const ProgrammingSelector = ({
...
+    }
+    // Handle string format: "Action"
+    if (genre != null) {
+      return String(genre);
+    }
+    return '';
+  };
+
+  const renderGenreChoices = () => {
+    // Handle both API response formats: { Items: [...] } or direct array
```
chrisbenincasa, line 471: IMO we should only have one definitive type here. It is defined at the API level. In this case we're returning the whole JF response object, so I think we should eliminate the need to handle the direct array approach

web/src/components/channel_config/ProgrammingSelector.tsx
chrisbenincasa: I think this will greatly simplify the logic below as well

web/src/hooks/jellyfin/useJellyfinApi.ts
```diff
@@ -27,6 +27,22 @@ export const useJellyfinUserLibraries = (
...
+  enabled: boolean = true,
+) => {
+  const apiClient = useTunarrApi();
+
+  return useQuery({
+    queryKey: ['jellyfin', mediaSourceId, 'libraries', libraryId, 'genres'],
+    queryFn: () =>
+      apiClient.getJellyfinGenres({ params: { mediaSourceId, libraryId } }),
+    enabled:
+      isNonEmptyString(mediaSourceId) && isNonEmptyString(libraryId) && enabled
```
chrisbenincasa, line 42: consider adding a `staleTime` of something relatively high here, since genres likely don't change that frequently per lib (could do 900_000 for 15m)

chrisbenincasa (changes requested):

types/src/jellyfin/index.ts
devbjackson: Have a look and let me know if this is what you meant.

web/src/components/channel_config/ProgrammingSelector.tsx
devbjackson: I believe I have addressed this now.

web/src/components/channel_config/ProgrammingSelector.tsx
devbjackson: I believe I have addressed this as well.

chrisbenincasa (approved):

server/src/external/jellyfin/JellyfinApiClient.ts
```diff
@@ -359,6 +359,27 @@ export class JellyfinApiClient extends BaseApiClient {
...
+      url: `/Genres`,
+      params: {
+        parentId,
+        userId: this.options.userId,
+        includeItemTypes,
+        recursive: 'true',
+      },
+    });
+
+    if (genresResult instanceof Error) {
```
chrisbenincasa, line 376: I just changed this recently to throw instead of return a Try - might be the cause of the build error

chrisbenincasa: I resolved this on my end - will merge once it's fixed!

Issue #746, opened 2024-09-05
Add channel action button menu to programming edit page


chrisbenincasa:
User should be able to get M3U link, etc from the programming edit page.

PR #1425, opened 2025-10-21
Add new channel options button across channel pages

markdavella:
Goal of this PR is to create a unified UX pattern for accessing channel options (copy ids, edit, programming, watch) across channel pages.  

This PR resolves https://github.com/chrisbenincasa/tunarr/issues/746

Updates:
- Created ChannelOptionsButton component that accesses the now renamed ChannelOptionsMenu
  - Now that we have this component we can adjust/tweak/improve the UX for this and have it propagate across the channel pages
- Created prop `hideItems` on ChannelOptionsButton with the type channelListOptions to allow us to control which options we list on which pages
- Added the options drop down to Edit, Add Programming, Channel Summary
- Removed some existing buttons to try and create consistency across channel pages on where these could be accessed
- Updated the 'edit' icon on Channels pages for consistency, it was using the "Settings" icon which we use elsewhere to indicate settings/options instead of edit.

Screenshots:

Channel Programming:
image

Channel Summary Page:
image

Watch Page:
image

Channels:
image

### DIFF
--- a/web/src/components/channels/ChannelOptionsButton.tsx
+++ b/web/src/components/channels/ChannelOptionsButton.tsx
@@ -1,0 +1,59 @@
+import type { channelListOptions } from '@/types/index.ts';
+import { Settings } from '@mui/icons-material';
+import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
+import { Button } from '@mui/material';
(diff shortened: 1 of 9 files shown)


chrisbenincasa:
Yes! Been wanting to do this for so long. Thank you!!

PR #1492, opened 2025-12-02
Wrote new Jellyfin client setup docs

kpupp:
I took the liberty of writing a Jellyfin guide. Not much different but definitely has some quirks that Plex doesn't! I think I did it right - if not let me know and I'll fix it!


chrisbenincasa:
Thanks! I'll have a look

PR #1559, opened 2025-12-23
fix: force 8-bit H.264 output to prevent High 10 profile playback issues

abra5umente:
## Problem

When transcoding 10-bit HEVC sources to H.264, libx264 preserves the 10-bit depth and outputs H.264 High 10 profile. This causes playback failures on most devices/players (including Plex) because hardware decoders rarely support 10-bit H.264.

## Solution

Add `-pix_fmt yuv420p` to the ffmpeg transcode command when using libx264, forcing 8-bit output regardless of source bit depth.

## Before
```
Stream: h264 (High 10), yuv420p10le
```

## After
```
Stream: h264 (High), yuv420p
```

## Testing
- Tested with 10-bit HEVC sources from Plex library
- Confirmed output is now 8-bit H.264 via ffprobe
- Plex Live TV now plays channels that previously failed

### DIFF
--- a/server/src/ffmpeg/builder/encoder/Libx264Encoder.ts
+++ b/server/src/ffmpeg/builder/encoder/Libx264Encoder.ts
@@ -15,6 +15,10 @@ export class Libx264Encoder extends VideoEncoder {
 
   options(): string[] {
     const opts = [...super.options()];
+    // Force 8-bit output (yuv420p) to ensure compatibility with most players/devices.
+    // Without this, libx264 may output 10-bit H.264 (High 10 profile) which many
+    // hardware decoders cannot handle.
+    opts.push('-pix_fmt', 'yuv420p');
     if (isNonEmptyString(this.videoPreset)) {
       opts.push('-profile:v', this.videoPreset);
     }


chrisbenincasa:
Thanks for the PR! I appreciate you digging into this — 10-bit H.264 High 10 output is definitely a real problem for playback compatibility.

  However, Tunarr already handles this at the pipeline builder layer rather than the encoder layer, so this change isn't needed and would actually be the wrong place to fix it:

  1. Pixel format is a pipeline concern, not an encoder concern. Every pipeline builder (software, VAAPI, NVIDIA, QSV, VideoToolbox) has a setPixelFormat() method that compares the current frame state against the
  desired state and emits -pix_fmt (via PixelFormatOutputOption) when conversion is needed. Encoders are only responsible for codec-level flags (-c:v, -preset:v, -profile:v).
  2. The desired state already defaults to yuv420p. Every FrameState construction in FfmpegStreamFactory.ts sets pixelFormat: new PixelFormatYuv420P(), so the pipeline builder should already detect a 10-bit source
  and emit the conversion.
  3. Adding it in the encoder would cause issues: it duplicates the flag when the pipeline also emits it, and it bypasses the pipeline's frame state tracking so downstream steps wouldn't know the pixel format
  changed.

  If you're still seeing 10-bit H.264 output despite this, there may be a bug in a specific pipeline path — I'd be happy to investigate if you can share your FFmpeg/transcoding settings and the stream details
  (codec, resolution, bit depth) of the source that produced the problem.

PR #1577, opened 2026-01-03
fix: add missing build dependency for meilisearch install

nstjuliana:
When running dev build instructions from README.md, I encountered errors when running `pnpm turbo dev`.

The install-meilisearch script imports from @tunarr/shared, but the task was missing this dependency in the Turbo graph. This caused a race condition where the script would attempt to run before the shared package was built.

This change adds `@tunarr/shared#build` to the `dependsOn` array for `install-meilisearch` to ensure the correct build order.

### DIFF
--- a/server/turbo.json
+++ b/server/turbo.json
@@ -1,41 +1,74 @@
 {
   "$schema": "https://turbo.build/schema.json",
-  "extends": ["//"],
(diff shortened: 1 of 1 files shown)


chrisbenincasa:
Thank you!

PR #1579, opened 2026-01-04
fix: populate subtitle language metadata from ffprobe

nstjuliana:
- Add languageCodeISO6392 field to subtitle stream details in FfprobeStreamDetails
- Add fallback to language field in FileSystemScanner when ISO code is missing

Previously, subtitle streams were missing the languageCodeISO6392 field which prevented subtitle language preferences from working correctly.


chrisbenincasa:
Thank you!

PR #1580, opened 2026-01-04
feat: enable direct burning of embedded text subtitles

nstjuliana:
- Remove pre-extraction requirement for text-based embedded subtitles
- Add stream index support to SubtitleFilter using subtitles=path:si=index syntax

This allows embedded text-based subtitles (like subrip) to be burned directly into the video stream without requiring pre-extraction to cache files.

### DIFF
--- a/server/src/ffmpeg/SubtitleStreamPicker.ts
+++ b/server/src/ffmpeg/SubtitleStreamPicker.ts
@@ -114,18 +114,9 @@ export class SubtitleStreamPicker {
           continue;
         }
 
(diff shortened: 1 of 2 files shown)


chrisbenincasa:
Appreciate the contribution - I left some context in a comment here, but I think embedded text subtitle stream extraction is a necessary evil for the time being.

chrisbenincasa:
Going to close this out for now, happy to discuss further over chat (Matrix or Discord)

PR #1601, opened 2026-01-19
Add endpoint to search by external id

Erotemic:
Adds a new API endpoint to search Tunarr content by a raw external id (the `external_key` stored on external-id rows), without requiring the `{source_type}|{source_id}|{external_key}` composite format.

### What’s included

* **New endpoint:** `GET /programming/external-id/search`

  * Required query: `externalId=<raw id>`
  * Optional filters: `sourceType`, `mediaSourceId`, `externalSourceId`
  * Paging: `limit`, `offset`
  * Toggles: `includePrograms`, `includeGroupings` (both default true)
  * Response returns matching `programs` and/or `groupings` plus totals.

* **DB migration:** adds indexes to keep the lookup fast:

  * `program_external_id(external_key)`
  * `program_grouping_external_id(external_key)`
  * Uses `CREATE INDEX IF NOT EXISTS` and the existing migration util for consistency.

Paging is likely not needed in normal use-cases, but added because its technically possible. 

Haven't tested yet, just at a stopping point.

### DIFF
--- a/server/src/api/programmingApi.ts
+++ b/server/src/api/programmingApi.ts
@@ -97,6 +97,19 @@ const BatchLookupExternalProgrammingSchema = z.object({
     }),
 });
 
+const SearchByRawExternalIdQuerySchema = z.object({
+  externalId: z.string().min(1).describe('Raw external id to match external_key'),
+
+  // Optional narrowing
+  sourceType: z.string().optional().describe('Optional source type filter (plex/jellyfin/emby)'),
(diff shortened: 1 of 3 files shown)


chrisbenincasa:
Thank you! I took a slightly different approach with #1607 and use the existing separate table + a join. This allows us to expand out the endpoint to include querying by IMDB/TMDB/etc in the future!

PR #1606, opened 2026-01-20
feat: add media_source_name and library_name search fields

### LINKED ISSUE #1524: Add possibility to scope searches by libraries
Some of the smart collections I want to create are dependent on the ability to scope searches to specific libraries.

For example I have libraries for regular movies and one for TV-Movies (movies spun off from tv-shows), I also have one library for Norwegian movies and one for Norwegian TV shows.

So for example here are some smart collections I would love to achieve:

- All movies from the "Norwegian movies" library only
  - `library:"Movies - Norwegian"` 
- All TV shows from the "Norwegian tv shows" library only
  - `library:"TV - Norwegian"` 
- All movies and TV shows from each of their respective libraries together
  - `library:"Movies - Norwegian" OR library:"TV - Norwegian"` 

And of course it _should_ hopefully still be possible to use all other search functions in combination, like `(library:"Movies - Norwegian" AND genre:Comedy) OR (library:"TV - Norwegian" AND genre:Comedy)`

(I used library names as examples, but using IDs would probably be better... or maybe both? 🤔 )

chrisbenincasa:
Closes #1524

### DIFF
--- a/server/src/api/mediaSourceApi.ts
+++ b/server/src/api/mediaSourceApi.ts
@@ -34,6 +34,7 @@ import type { MarkOptional, StrictExtract } from 'ts-essentials';
 import { match, P } from 'ts-pattern';
 import { v4 } from 'uuid';
 import z from 'zod/v4';
+import { DeleteMediaSourceCommand } from '../commands/media_source/DeleteMediaSourceCommand.ts';
 import { container } from '../container.ts';
 import type { MediaSourceWithRelations } from '../db/schema/derivedTypes.js';
 import { EntityMutex } from '../services/EntityMutex.ts';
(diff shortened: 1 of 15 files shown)


k1rsch:
Both "media_source_name" and "library_name" aren't working correctly in the search field. If there should be a match I get no results. If there isn't a match I get results from all Media Sources and Libraries.

chrisbenincasa:
@k1rsch What version are you running

k1rsch:
1.2.0-dev.5

chrisbenincasa:
Should be fixed in https://github.com/chrisbenincasa/tunarr/commit/850440d319239cc44044151dab7dc4e58c74248c which will be in the next pre-release

PR #1623, opened 2026-01-27
fix: potential fixes for HLS streaming

chrisbenincasa:


fancypantalons:
So I strongly suspect these changes introduced a streaming regression. I can consistently replicate it as follows:

- On device one, start a streaming session for a channel through the browser
- On device two, start a second streaming session for the same channel
- Terminate the stream for device two
- Playback hangs on device one
- Restart the stream on device two and playback resumes on device **one**

This behaviour is *not* observed is no second device is in the mix.

Looking at the network traffic versus what's on disk, I see an updated stream manifest written to disk but in the browser the manifest is frozen/unchanging.

I suspect this is because `_minByIp` never sees entries retired from the map. The result is that an old, defunct session can "anchor" the start of the manifest. And because the manifest is limited to a certain number of segments, the manifest freezes.

When device two restarts its stream, its entry in `_minByIp` gets updated, the manifest is no longer anchored, and then device one starts seeing updated manifests once again.

chrisbenincasa:
@fancypantalons - thanks for the thorough write-up. I will look into this. I think you are likely correct.

PR #1631, opened 2026-02-01
fix: isolate slot iteration state in time slots

### LINKED ISSUE #616: Time slot rerun functionality
Add the ability to schedule 'reruns' directly in the time slot editor. 

We'll have to think a bit about the semantics here and how we'd present it in the UX. The basic idea is essentially assigning multiple time slots to the same show and syncing them, e.g. "At 10am and 4pm, play the Simpsons"

chrisbenincasa:


chrisbenincasa:
Superseded by https://github.com/chrisbenincasa/tunarr/pull/1838

Issue #1637, opened 2026-02-06
[Feature] Add Option to Proxy Client Browser Image Calls Through Tunarr


PhantomOffKanagawa:
**This solves:** When using an internal url for a source (i.e. on k3s `http://jellyfin.media-automation.svc.cluster.local:80` for jellyfin) the client browser makes requests for posters and art to that internal url. This means all image requests fail and the site uses placeholders exclusively

**A solution**: Add a flag or setting to proxy these image calls through Tunarr which has access to the source via that url.

**Additional solutions**
- Add a setting on the source for an "External URL" used in the xmltv, image requests, etc. wherever a service external to Tunarr would access the source

**Additional context**

[full-redacted-logs.log](https://github.com/user-attachments/files/25115584/full-redacted-logs.log)

<img width="599" height="736" alt="Image" src="https://github.com/user-attachments/assets/b41464c5-2bbc-467a-b962-944ba90329a7" />

<img width="838" height="888" alt="Image" src="https://github.com/user-attachments/assets/731e8b5a-606d-4a2e-977a-e180d8e7aef8" />

<img width="838" height="888" alt="Image" src="https://github.com/user-attachments/assets/749badf5-7b89-4349-9dba-4ad34f1b89c7" />

Started in [this discord help thread](https://discord.com/channels/1254564006123802805/1468763897107320927)

SystemIdleProcess:
Could images be stored using relative path and use request host:port? I'm still new to reverse proxy but am attempting to set one up so that I can use an origin cert and SSL.

chrisbenincasa:
This is included in 1.2 using the `TUNARR_PROXY_ARTWORK` env var. I plan to make it the default behavior in 1.3

PR #1638, opened 2026-02-06
feat: env var for proxying images through tunarr

PhantomOffKanagawa:
Related to issue: [https://github.com/chrisbenincasa/tunarr/issues/1637](https://github.com/chrisbenincasa/tunarr/issues/1637)

Adds the environment variable `TUNARR_PROXY_ARTWORK` which when "true" makes the image calls on the client browser proxy through Tunarr. This is helpful for cases where a Tunarr source connects via an internal address that the client's browser can't access.

I reused the proxy code for `/programs/:id/thumb` from `programmingApi.ts:710-746`

### DIFF
--- a/server/src/api/creditsApi.ts
+++ b/server/src/api/creditsApi.ts
@@ -195,7 +195,7 @@ export class CreditsApiController {
               }
 
               return res
-                .status(proxyRes.status)
+                .status(200)
                 .headers(headers)
                 .send(proxyRes.data);
             } catch (e) {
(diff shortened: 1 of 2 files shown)


chrisbenincasa:
Will check the code shortly: two quick things 

1. Let's prefix the env var with TUNARR_ to match elsewhere
2. Can you change the base branch for the PR to dev

PhantomOffKanagawa:
Sorry, my github notifications are super flaky so I didn't see this till yesterday. Thank you for changing the base branch, and I already had it prefixed with `TUNARR_` but I copied the wrong thing into my original post so I edited it to fix it.

PhantomOffKanagawa:
Looking at the build errors now. I've built it locally so I think they are essentially just type linting errors but I will fix them

PhantomOffKanagawa:
Should be fixed

PR #1654, opened 2026-02-18
fix(plex): map credit markers to 'outro' chapter type

awernick:
## Summary

Plex credit markers (`marker.type === 'credits'`) were incorrectly mapped to chapter type `'intro'` instead of `'outro'` when importing program metadata. This made real intro markers indistinguishable from credit markers in the database.

## Changes

- **PlexApiClient.ts**: Map Plex `credits` marker type to `'outro'` instead of `'intro'`
- **Migration 0039**: Delete ambiguous `'intro'` chapters for Plex-sourced programs and null their `canonicalId` so the next library scan re-imports with correct types

## Notes

- The migration is destructive by necessity. Existing `'intro'` rows are a mix of real intros and mislabeled credits with no way to tell them apart. After running, a library re-scan re-imports all chapters correctly.
- Non-Plex programs (Jellyfin, Emby) are unaffected.
- The migration is idempotent.

### DIFF
--- a/server/src/migration/DirectMigrationProvider.ts
+++ b/server/src/migration/DirectMigrationProvider.ts
@@ -186,6 +186,10 @@ export class DirectMigrationProvider implements MigrationProvider {
           migration1768825617: makeKyselyMigrationFromSqlFile(
             './sql/0037_orange_bromley.sql',
           ),
+          migration1771459201_FixMislabeledCreditMarkers:
+            makeKyselyMigrationFromSqlFile(
+              './sql/0039_fix_mislabeled_credit_markers.sql',
+            ),
         },
         wrapWithTransaction,
       ),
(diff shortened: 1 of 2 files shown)


chrisbenincasa:
Thanks for the fix! I don't think this requires a DB migration fix (and I don't usually do those on main branch PRs anyway). The data can be fixed with a force scan of affected libraries because credits are wholesale removed / inserted each time (like a set operation) https://github.com/chrisbenincasa/tunarr/blob/main/server/src/db/ProgramDB.ts#L1819

awernick:
@chrisbenincasa Thanks for the reply! 

I initially through a force rescan of the library would be sufficient but I believe there’s logic that prevents us from re-fetching / re-storing the markers unless the canonical ID has changed (sorry I’m probably butchering some of these terms, I looked at them way too late last night).

I’ll double check this though and remove the migration if it ends up not being necessary!

chrisbenincasa:
A force rescan skips the checks of whether the source has changed

chrisbenincasa:
Canonical check bypassed here: https://github.com/chrisbenincasa/tunarr/blob/main/server/src/services/scanner/MediaSourceMovieLibraryScanner.ts#L106-L107

chrisbenincasa:
Ping me when the migration stuff is removed here and I'll merge!

chrisbenincasa:
Fixed this with c03e7bbb

PR #1655, opened 2026-02-18
feat(stream): skip end credits using Plex outro markers

awernick:
Adds a per-channel "Skip End Credits" toggle. When enabled, programs stop playback when credits begin (using Plex credit markers already stored as chapter data) and the next program starts immediately. The existing flex/filler system fills any remaining schedule gaps naturally.

### Changes

- **Channel schema + migration 0038**: Add `skipCredits` boolean column, wired through DB converters, API types, and OpenAPI spec
- **StreamProgramCalculator**: Cap `streamDuration` at outro start time.
  If the viewer tunes in during credits, advance to the next program
- **Web UI**: Toggle under new Playback section in channel Streaming settings
- **Tests**: 5 new tests (duration cap, mid-credits advance, no-outro passthrough, disabled toggle, filler item immunity)

### Notes

- Depends on correct `outro` chapter types from #1654
- Only affects programs with Plex credit markers. Programs without markers play normally regardless of the toggle.
- Defaults to off. No behavior change for existing channels.

### DIFF
--- a/server/src/db/ChannelDB.test.ts
+++ b/server/src/db/ChannelDB.test.ts
@@ -117,6 +117,7 @@ function createSaveableChannel(
     streamMode: 'hls' as const,
     transcodeConfigId,
     subtitlesEnabled: false,
+    skipCredits: false,
     ...overrides,
   };
 }
--- a/server/src/db/ChannelDB.ts
+++ b/server/src/db/ChannelDB.ts
@@ -201,6 +201,7 @@ function updateRequestToChannel(updateReq: SaveableChannel): ChannelUpdate {
     transcodeConfigId: updateReq.transcodeConfigId,
     streamMode: updateReq.streamMode,
     subtitlesEnabled: booleanToNumber(updateReq.subtitlesEnabled),
+    skipCredits: booleanToNumber(updateReq.skipCredits),
   } satisfies ChannelUpdate;
 }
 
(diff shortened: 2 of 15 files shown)


chrisbenincasa:
This is a great start! I think there's a missing piece here though - specifically, by skipping outros, we're inherently changing the duration of programs that have the outro markers. This will affect guide building. By only skipping the outros in the stream calculator, over time the stream will skew from the built guide as the former sees the "shorter" duration while the guide will show the full duration of the item. Same goes for the part in the PR when you tune in during credits (effectively skipping ahead in the schedule). 

In general Tunarr tries hard to keep the EPG synced with what is actually going to play when you tune in. If we want to support this, we'll need to detect outros at guide building time (this will inherently be a synced library only feature too, since that data isn't available for items that are from unsynced libraries) and build it right into the schedule. This also means the guide needs to be rebuilt whenever the channel setting changes; I'm working on a change that should make that part a bit easier to detect.

SystemIdleProcess:
I'm sure this is way more difficult but you know how during credits the channel would put the credit scroll in a smaller window in the upper right (and even speed them up) and role promos for the rest of the "main" screen. That'd be a fun addition.

chrisbenincasa:
Gonna close this as stale. Would be curious to implement this more fully, but I don't have the cycles right now to do it.

PR #1659, opened 2026-02-22
feat: experimental tonemapping support for CUDA hwaccel

chrisbenincasa:


chrisbenincasa:
cc: @coreyjv  - here's a basic CUDA implementation, based off of your changes for VAAPI. We can keep iterating here for other paths. Will merge this once your changes are in, I still have some testing and cleanup to do.

coreyjv:
Sounds good @chrisbenincasa. I'll hopefully finish up in the next couple of day or so. Thanks for basing off of what I got and the refactors you've done make sense.

chrisbenincasa:
No rush! We may be able to merge as is and add tests in a follow up.

PR #1674, opened 2026-02-26
docs: add other video tags info and example structure

daryl249:
Quick explanation so others trying this don't get lost the way I did.


chrisbenincasa:
Thanks!

PR #1675, opened 2026-02-26
fix: remove deleted channels from guide cache

coreyjv:
Fixes an issue where deleted channels were returned in XMLTV data until the server was restarted.

### DIFF
--- a/server/src/api/channelsApi.ts
+++ b/server/src/api/channelsApi.ts
@@ -349,11 +349,9 @@ export const channelsApi: RouterPluginAsyncCallback = async (fastify) => {
       await req.serverCtx.m3uService.regenerateCache();
 
       try {
-        GlobalScheduler.getScheduledJob(UpdateXmlTvTask.ID)
-          .runNow()
-          .catch((err) => logger.error(err, 'Error regenerating guide'));
+        await req.serverCtx.guideService.removeCachedChannel(channel.uuid);
       } catch (e) {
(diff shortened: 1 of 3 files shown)


chrisbenincasa (changes requested):
I think it would also be good to hook into the channel DELETE API endpoint, remove it from the cached guide, and rewrite the XMLTV (no regeneration needed). This change is fine, the XMLTV will be eventually consistent with the actual state in Tunarr, but there are already mechanisms for breaking the cache and rewriting XMLTV w/o generation, so I think we should just hook into that as well while we're doing the change.

coreyjv:
Just to make sure I understand you're fine with the proposed changes but you also want me to take a look at the delete channel API as well or are you suggesting that I rework this to just deal with the delete channel API?

coreyjv:
> but there are already mechanisms for breaking the cache and rewriting XMLTV w/o generation

Where can I find those examples?

coreyjv:
Also with the delete endpoint additions do you want to replace the entire XmlTv update task (noted [here](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/channelsApi.ts#L351-L357)) with the targeted cache update and XMLTv write?

chrisbenincasa:
Thanks for this! This has been a long standing bug that I'm happy to get fixed :)

coreyjv:
> Thanks for this! This has been a long standing bug that I'm happy to get fixed :)

No problem!

PR #1676, opened 2026-02-26
Embiggen tag selector

daryl249:
Make the "Value" selectorbox larger so it's easier to see file paths when using this with "Tags" field.


chrisbenincasa (changes requested):
Looks like this needs a rebase because it includes commits from https://github.com/chrisbenincasa/tunarr/pull/1674

daryl249:
Whoops.  Saw you're using a standard for commit messages and fixed that up too.

chrisbenincasa (approved):

PR #1678, opened 2026-02-27
Smart grouping for marathon style channels

stasadance:
I wanted to make a channel where I had "marathon" style playback so different movie series would get grouped together in chronological release order. The only way I saw this working was with the tags attribute on each movie, in my case with plex they include enough information to link collections together as each movie series has the same collection tag.

I'm not too familiar with this repo yet but was able to add this to the slots editor functionality with help of claude. Please excuse any mistakes, will work on any suggestions and improvements as necessary. Would love for help in testing and making this feature better, I have pushed a tagged build on my fork `ghcr.io/stasadance/tunarr:smart-grouping`

image

 ## Summary
   - Adds a **Group by Tag** toggle to movie and smart collection slots, enabling marathon-style playback where
   programs sharing a tag (e.g., "Harry Potter Collection") play back-to-back
   - Groups are ordered by the slot's existing order setting (shuffle, alphabetical, chronological, etc.), while
   programs within each group always play in chronological/release order
   - Configurable options for untagged programs (include as individuals or exclude) and multi-tag behavior (first
   match or all matching groups, deduplicated)

   ## Changes
   - **Types**: New `SlotGroupBySchema` with `type`, `ungrouped`, and `multiTagBehavior` fields; added optional
   `groupBy` to movie and smart-collection slot schemas
   - **Server DB**: Added tag relation loading to `getChannelAndPrograms` and `getProgramGroupingDescendants` so tags
    are available during scheduling
   - **Server Scheduling**: New `groupProgramsByTag()` pre-processing step that buckets programs by tag, sorts within
    groups chronologically, orders groups by slot order, and deduplicates across groups
   - **Web UI**: New `SlotGroupByFormControl` component with toggle switch and dropdown options, added to movie slot
   and smart collection slot editors

### DIFF
--- a/server/src/services/scheduling/slotSchedulerUtil.ts
+++ b/server/src/services/scheduling/slotSchedulerUtil.ts
@@ -14,6 +14,7 @@ import type {
   BaseSlot,
   FillerProgrammingSlot,
   SlotFillerTypes,
+  SlotGroupBy,
 } from '@tunarr/types/api';
 import { FillerTypes } from '@tunarr/types/schemas';
 import type { Duration } from 'dayjs/plugin/duration.js';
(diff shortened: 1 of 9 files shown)


types/src/api/CommonSlots.ts
```diff
@@ -43,6 +43,14 @@ export const Slot = z.object({
   filler: z.array(SlotFiller).optional(),
 });
 
+export const SlotGroupBySchema = z.object({
+  type: z.literal('tag'),
```
chrisbenincasa: This is only really necessary if we are using a discriminated union and there are other types of "group by"s we need to think about. If you think there will be more in the future, we could preemptively make this a discriminated union -- otherwise I think we can omit this field entirely.

server/src/services/scheduling/slotSchedulerUtil.ts
```diff
@@ -474,11 +494,128 @@ export function slotFillerIterators(
...
+ *
+ * When order is 'shuffle', groups are shuffled using the provided Random.
+ * When order is 'alphanumeric', groups are sorted A-Z by tag name.
+ * When order is 'next' or 'chronological', groups are sorted by earliest year.
+ * When order is 'ordered_shuffle', groups are sorted then rotated randomly.
+ */
+export function groupProgramsByTag(
+  programs: SlotSchedulerProgram[],
+  groupBy: SlotGroupBy,
+  order: string,
```
chrisbenincasa: Unfortunately we don't have a named type here, but let's prefer `BaseSlot['order']` or something similar for tighter type checking.

chrisbenincasa (changes requested):
Hey! Thanks for the PR. I generally like the idea, and it is something that has been asked for in different ways by users. I do think that there are some gaps in behavior and higher - level questions I have, in addition to some improvements in the code itself to match the style of Tunarr and reuse existing functions, etc for consistent behavior. I've left a first round of comments.

chrisbenincasa:
Friendly ping :)

PR #1702, opened 2026-03-12
docs: Replace errant "movies" reference with "other videos" in severa…

daryl249:
…l places and fix plurals in others


chrisbenincasa:
Thanks!

PR #1705, opened 2026-03-13
feat: dolby vision tonemapping for vaapi

coreyjv:
Adds detection and tonemapping support for Dolby Vision to VAAPI and software pipelines.

### DIFF
--- a/server/src/ffmpeg/builder/MediaStream.test.ts
+++ b/server/src/ffmpeg/builder/MediaStream.test.ts
@@ -1,5 +1,8 @@
 import { VideoStream } from '@/ffmpeg/builder/MediaStream.js';
-import { PixelFormatYuv420P } from '@/ffmpeg/builder/format/PixelFormat.js';
+import {
+  PixelFormatYuv420P,
+  PixelFormatYuv420P10Le,
+} from '@/ffmpeg/builder/format/PixelFormat.js';
 import { FrameSize } from '@/ffmpeg/builder/types.js';
 
 describe('MediaStream', () => {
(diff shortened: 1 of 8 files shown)


chrisbenincasa:
Is this missing the VAAPI part? I only see the software part.

coreyjv:
> Is this missing the VAAPI part? I only see the software part.

Nope! `tonemap_vaapi` does not deal directly with Dolby Vision. So the part that benefits VAAPI here is the detection of DV media. It will then filter that through the same HDR10 process using the base later. In essence it allows more content to be tonemapped in VAAPI than the original implementation.

chrisbenincasa:
Ah ok, so there's nothing really VAAPI specific here? This would theoretically apply to any hwaccel that uses the same helper to detect whether to tonemap?

coreyjv:
> Ah ok, so there's nothing really VAAPI specific here? This would theoretically apply to any hwaccel that uses the same helper to detect whether to tonemap?

Ya, I would say that's true. Sorry for any confusion here. There is additional work in the NVidia/Vulkan space to use libplacebo to properly tonemap DV metadata.

chrisbenincasa:
What is required on the libplacebo side? Quick search makes it sound like libplacebo automatically detects metadata from incoming streams, applies necessary transforms, and then strips it from the output? 
image
 disclaimer: these are ffmpeg8 docs

coreyjv:
So I think this is dependent on the libplacebo version it seems that an additional option needs to be provided to the `libplacebo` filter ala `dolbyvision=1` or `dolbyvision=true`. That's it. 

I'm happy to push that commit an we can just check-off DV across all pipelines for now :D

coreyjv:
Although I just looked at the ffmpeg 7 filter docs and that option seems to be there too. So we may be all set on the Nvidia/Vulkan pipeline.

So if we merge this I think we get DV for everything. We could do that for now and get feedback from folks.

chrisbenincasa:
Sounds good, let's give it a go. Code LGTM. I think this is the last merge before 1.2!

PR #1732, opened 2026-03-21
fix: lazy initialize logger so logging works

coreyjv:
Logging wasn't working because they were using a static field initializer which ran before the logger factory was initialized. Switch to a lazy initialization so it's initialized on first use and logs with the defined root logger.

### DIFF
--- a/server/src/api/ApiProgramConverters.ts
+++ b/server/src/api/ApiProgramConverters.ts
@@ -32,11 +32,12 @@ import type {
 } from '../services/MeilisearchService.ts';
 import type { Maybe, Nullable } from '../types/util.ts';
 import { isNonEmptyString } from '../util/index.ts';
+import type { Logger } from '../util/logging/LoggerFactory.ts';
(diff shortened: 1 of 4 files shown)


server/src/api/ApiProgramConverters.ts
```diff
@@ -36,9 +36,12 @@ import { LoggerFactory } from '../util/logging/LoggerFactory.ts';
 import { titleToSortTitle } from '../util/programs.ts';
 
 export class ApiProgramConverters {
-  private static logger = LoggerFactory.child({
-    className: ApiProgramConverters.name,
-  });
+  private static _logger: ReturnType<typeof LoggerFactory.child> | undefined;
```
chrisbenincasa: I think an ideal world we'd have these all converted to use the inject pattern to get the logger. Also in general it's easier on the TS compiler to just use the return type rather than making it infer via ReturnType

coreyjv: Sounds good. I’m not as familiar with that framework used here but I’ll take a crack at it. Thanks for the feedback!

PR #1735, opened 2026-03-24
chore: ignore files generated from JetBrains IDEs

coreyjv:


chrisbenincasa:
Let's change the commit type from `feat` to `chore` (or `build`) - this doesn't actually change functionality and shouldn't trigger a version change.

coreyjv:
Fixed!

PR #1755, opened 2026-03-28
guard against non-finite values in stream play time cache

RickConsole:
`recordProgramPlayTime` in `ChannelCache.ts` can write NaN to the persistent stream cache when lineupItem.duration is undefined or when upstream time calculations produce non-finite values. Once written, JSON.stringify(NaN) becomes null, which fails Zod validation on the next cache read. This causes a bug that I've been facing recently which corrupts the entire guide data of all channels. 

This fix adds a Number.isFinite() check before writing to the cache. If the computed play time is non-finite, the write is skipped and a warning is logged instead of silently corrupting stream-cache.json. Users with an already-corrupted cache can delete stream caches to reset.

### DIFF
--- a/server/src/stream/ChannelCache.ts
+++ b/server/src/stream/ChannelCache.ts
@@ -101,15 +101,27 @@ export class ChannelCache implements IStreamLineupCache {
       remaining = lineupItem.duration - (lineupItem.startOffset ?? 0);
     }
 
+    const playTime = t0 + remaining;
+
+    if (!Number.isFinite(playTime)) {
+      this.logger.warn(
+        'Refusing to write non-finite playTime for channel %s (t0=%s, remaining=%s)',
+        channelId,
+        t0,
+        remaining,
+      );
+      return;
+    }
+
     if (lineupItem.type === 'program') {
       const key = this.getKey(channelId, lineupItem.program.uuid);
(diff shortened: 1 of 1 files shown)


chrisbenincasa:
Thanks! The stream cache file isn't actually read back anymore - its functionality has been converted to the program play history table. So while this fix is good and will work, I don't think there will be any effect.

RickConsole:
No problem! And that's interesting. Was this a recent fix? My guide data was corrupted several days ago due to this NaN issue.

EDIT - Ahh i see database functionality was added recently. My apologies!

chrisbenincasa:
The change is not that recent. There were two parts to it. First I removed the code that read from the stream (channel) cache in e5e5b1510660149ebd6404ba6bc87ad6b500403a, this was 5 months ago. Then fe2a460546f0c2762ade0c2c18b654ac9139fc6e added the sql table backed replacement. This was 2 months ago and released in 1.2. The sql table replacement is used to power the filler picker only. Nothing has read from the stream cache for a few months and even when it did, it was not used in guide data calculation but rather at stream time.

chrisbenincasa:
Closing this out since I've removed all of this code in https://github.com/chrisbenincasa/tunarr/commit/61b2839a39d92f841995670648d9c2f120502fef

PR #1768, opened 2026-03-30
fix: improve user experience of channel overview on mobile

coreyjv:
Added:
- Scrollable tabs to view media
- Changed heading sizes
- Scrollable tabs on program details
- Fixed bullet being added to end incorrectly
- Improved program info ux
- Fixed an issue with `useState` inside of `useEffect`

Screenshots

| Before | After |
| ------ | ----- |
| before-1 |  after-1 |
| before-2 | after-2 |
| before-program-details | after-program-details |

### DIFF
--- a/web/src/components/ProgramMetadataDialogContent.tsx
+++ b/web/src/components/ProgramMetadataDialogContent.tsx
@@ -14,7 +14,7 @@ import type { Dayjs } from 'dayjs';
 import dayjs from 'dayjs';
 import { capitalize } from 'lodash-es';
 import type { ReactEventHandler } from 'react';
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useMemo, useRef, useState } from 'react';
 import { useSettings } from '../store/settings/selectors.ts';
 import ProgramInfoBar from './programs/ProgramInfoBar.tsx';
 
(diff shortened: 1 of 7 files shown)


chrisbenincasa:
Wondering if we can get the options button to be on the same row as the Channel name header for mobile (and maybe even on desktop?) On mobile it may look better as an icon button (no text) too. Thoughts?

coreyjv:
I thought about it but decided against it on mobile because I figured it’d cut the name off or force us to wrap.

I can try the icon experience but that’s still going to take up valuable space (~10%). What do you think?

Could we merge this change and I can take a look at options?

chrisbenincasa:
Good point - a few options I can think of:

1. Put the icon or button above the channel text. I feel like the gap in its current form looks a little wonky
2. Use a FAB: https://mui.com/material-ui/react-floating-action-button/ and/or optional speed dial with the actions: https://mui.com/material-ui/react-speed-dial/. There's a handful of actions so might be worth simplifying for mobile (or just making it FAB that goes to edit)

coreyjv:
I took option #1 and pushed up some changes. Here's what it looks like now:
channel-icon-stack

chrisbenincasa:
Love it!

PR #1813, opened 2026-04-18
fix: improve user experience of library overview on mobile

coreyjv:
Replace MRT table with card layout on small viewports and scale the page heading down to h5 on mobile.

| Before | After |
|--------|--------|
| before | after |

### DIFF
--- a/web/src/components/MediaSourceLibraryTable.tsx
+++ b/web/src/components/MediaSourceLibraryTable.tsx
@@ -6,12 +6,18 @@ import {
 } from '@mui/icons-material';
 import {
   Box,
+  Card,
+  CardContent,
+  Chip,
   IconButton,
   Link as MuiLink,
   Paper,
+  Stack,
   TableContainer,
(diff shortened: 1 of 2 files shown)


chrisbenincasa:
Looking awesome. Wonder if just using icons to indicate the source type would be sufficient? And then we'd be able to cut down on horizontal space of that second line even more

coreyjv:
> Looking awesome. Wonder if just using icons to indicate the source type would be sufficient? And then we'd be able to cut down on horizontal space of that second line even more

Do we have pre-existing icons that align with the source types? I could also move the scan time to a third line for now so it doesn’t cause that overlap.

chrisbenincasa:
Yes there should be SVGs for all remote sources and can find something appropriate for local in the material icons. There are some examples around the codebase

coreyjv:
Here ya go!
after-2

chrisbenincasa:
Looks awesome!

Issue #1797, opened 2026-04-12
[Feature Request] Able to remove the default icon from a channel


mcfarlandaaron1:
So I am wondering if there is a way to remove the channel icon from channels you don't have a specific logo? The problem I am running into I am using in the app Channels if you don't have specific icon it uses the default here.  The channel name doesn't show just the icon which make a challenge to figure out what channel is which in their EPG.  I have not found a way to just using the name and not to use the icon.  Is there a way I can remove the icon for channels if you don't have a logo?  Thank you.

Here is the message link discussing this in discord with a couple of image examples to understand what I am requesting. https://discord.com/channels/1254564006123802805/1254564006123802808/1492721256892727307

PR #1817, opened 2026-04-18
feat: add support to remove icon from channel

coreyjv:
Implements #1797.

### DIFF
--- a/server/src/api/channelsApi.ts
+++ b/server/src/api/channelsApi.ts
@@ -37,6 +37,8 @@ import {
 } from '@tunarr/types/schemas';
 import dayjs from 'dayjs';
 import duration from 'dayjs/plugin/duration.js';
+import { globalOptions } from '@/globals.js';
+import { deleteIfLocalAndCleared } from '@/util/iconUtil.js';
 import {
   head,
   isEmpty,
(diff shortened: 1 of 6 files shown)


chrisbenincasa:
ditto here re: build errors; should be fixed upstream. just needs a rebase

coreyjv:
Should be all set.

chrisbenincasa:
IIUC i think the feature was also asking for the ability to clear the icon even from the default (i.e. have no icon). Not just clear an uploaded one?

coreyjv:
Oh that’s my mistake. Did we want to merge this and then open another PR for part 2?

coreyjv:
I think it's worth noting that to support _no icon_ the changes become a bit more extensive. It involves understanding three states (e.g. custom icon, default icon, no icon) and requires changes from database code (depending on how we want to do it) all the way to writing the TV Guide / XML data.

My recommendation at this point would be to merge this change and then determine if we want to proceed with supporting the three states.

chrisbenincasa:
Yea, it's definitely a little more complicated. When i saw the issue I envisioned it working like:

1. Add an optional field to the channel icon JSON schema (the DB column is JSON) for `useDefaultIconFallback` (or something similar) - this will default to true on undefined, something we can do at the zod schema level
2. Add a toggle on the UI which changes this field in the DB for the channel
3. Omit the property/attribute entirely for icon in the XMLTV when the bit is false

Fine to merge this as-is to keep the changes separate and contained.

coreyjv:
Ya, I prefer to merge and that way I can open a PR with this specific change which will allow us to have a more focused discussion. We’ll just consider this part 1 :)

PR #1828, opened 2026-04-21
fix: update NFO parsing to include studio field

### LINKED ISSUE #1827: [Bug]: Studio information not picked up from nfo files
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [Discord](https://discord.gg/scMJxnPXaB) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/chrisbenincasa/tunarr/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Tunarr; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.

### Description of the bug

Tunarr (vers 1.2.11) detail on a TV show or a Movie has a blank studio(s) field, despite the related nfo file having the studio(s) information listed. This prevents creating TV show channels such as ABC, NBC, etc., movie franchise collections, etc., or otherwise using Studio in a search function. Attached are the media detail for 2 representative items from Tunarr as well as the related nfo files in one zip file.

[tunarr_nfo_issue.zip](https://github.com/user-attachments/files/26935845/tunarr_nfo_issue.zip)

### Reproduction steps

Scan media (TV shows, movies), ensuring that the related nfo file contains Studio data, view the detail for the media item in Tunarr and the Studio(s) field is blank.

### What is the current _bug_ behavior?

Tunarr does not read the Studio(s) field data from nfo files.

### What is the expected _correct_ behavior?

Tunarr would read the Studio(s) data from nfo files.

### Tunarr version

Latest

### FFMPEG encoder type

None

### Specify full Edge version id

_No response_

### Specify Alpha release number

_No response_

### Specify version number

_No response_

### What browser did you experience the the problem on? Please provide if the issue is related to the Web UI and definitely not the server.

_No response_

### Deployment Type

Docker

### What operating system are you using?

Ubuntu

### Full server logs

```shell
No logs attached, instead please see attached zip file with respective nfo files and Tunarr media detail json
```

coreyjv:
Fixes #1827.

### DIFF
--- a/server/src/nfo/MovieNfoParser.test.ts
+++ b/server/src/nfo/MovieNfoParser.test.ts
@@ -14,6 +14,21 @@ describe('MovieNfoParser', () => {
     expect(output.movie.credits).length(2);
     expect(output.movie.director).length(1);
     expect(output.movie.actor).length(5);
+    expect(output.movie.studio).toHaveLength(1);
+    expect(output.movie.studio).toEqual(['Marv Films']);
+  });
+
+  test('parses basic movie file no studio', async () => {
+    const contents = (await readTestFile('movie-no-studio.nfo')).toString(
+      'utf-8',
+    );
+    const result = await new MovieNfoParser().parse(contents);
+    expect(result.isSuccess()).toBe(true);
+
+    const output = result.get();
+
+    expect(output.movie).toBeDefined();
+    expect(output.movie.studio).toBeUndefined();
   });
 
   test('parses NFOs without explicit fileinfo', async () => {
--- a/server/src/nfo/MovieNfoParser.ts
+++ b/server/src/nfo/MovieNfoParser.ts
@@ -12,6 +12,7 @@ const ArrayTags = [
   'movie.fileinfo',
   'movie.actor',
   'movie.uniqueid',
+  'movie.studio',
 ];
 
 export class MovieNfoParser extends BaseNfoParser<typeof MovieNfoContainer> {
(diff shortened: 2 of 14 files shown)


server/src/nfo/MovieNfoParser.test.ts
```diff
@@ -14,6 +14,8 @@ describe('MovieNfoParser', () => {
     expect(output.movie.credits).length(2);
```
chrisbenincasa, line 14: Can we add 2 new tests - one NFO w/o studio (assert empty array) and one with multiple (assert length and contents)

coreyjv: I can but without I believe should be `.toBeUndefined()`. The parsing doesn't return empty arrays by default for `.optional()` fields. The empty arrays are created in the respective scanners (you can see I follow the pattern there).

coreyjv:
I'm adding some tests for TV Episode parsing as well. Stand by.

server/src/nfo/MovieNfoParser.test.ts
chrisbenincasa: Oh that's fine, as long as it's some sort of empty

coreyjv: I think I met the intent with the changes then.

chrisbenincasa: I think it's fine to follow the pattern of the rest of the array fields. undefined vs empty array doesn't really matter at the storage level in this case, so either technically works.

PR #1830, opened 2026-04-22
feat: add three-state channel icon support (custom, default, none)

coreyjv:
Add `useDefaultIconFallback` boolean to `ChannelIcon` schema to support a third icon state — no icon at all — distinct from the default Tunarr logo. When false and path is empty, M3U tvg-logo, XMLTV icon, guide watermark overlay, and all UI display locations suppress the icon rather than falling back to the Tunarr logo.

Part 2 of #1797.


server/src/services/XmlTvWriter.ts
```diff
@@ -93,13 +94,15 @@ export class XmlTvWriter {
       ],
     };
 
-    if (channel.icon) {
+    const iconSrc = resolveIconUrl(channel.icon, '{{host}}/images/tunarr.png');
+    if (iconSrc != null) {
```
chrisbenincasa: [nit] `if (isNonEmptyString(iconSrc))` or even more simply `if (iconSrc)` if we know it'll always return null or non-empty

server/src/util/iconUtil.ts
```diff
@@ -27,6 +28,21 @@ export function extractLocalUploadFilename(url: string): string | null {
...
+ * - Default fallback enabled (useDefaultIconFallback === true, or icon missing): returns defaultUrl
+ * - No icon (path empty, useDefaultIconFallback === false): returns null
+ */
+export function resolveIconUrl(
+  icon: ChannelIcon | null | undefined,
+  defaultUrl: string,
+): string | null {
+  if (!icon) return defaultUrl;
+  if (isNonEmptyString(icon.path)) return icon.path;
+  return icon.useDefaultIconFallback !== false ? defaultUrl : null;
```
chrisbenincasa, line 43: [nit] generally prefer shorthand `!!icon.useDefaultIconFallback` - don't feel strongly though

chrisbenincasa:
This will need a rebase, I merged a lot of changes recently. There is also a new presubmit check for ensuring that there are no missing message translations as part of the PR. You'll need to run `pnpm turbo compile-messages` to extract + generate any missing message translation code. I'm working on how to better automate this.

server/src/util/iconUtil.ts
coreyjv: This change actually fails an ESLint validation. I'll keep it as is if you're fine with that.

coreyjv:
Addressed comments. Rebased and added messages as requested.

PR #1865, opened 2026-05-07
Add optional basic auth for protected routes

Trifocals3537:
## Summary

Adds optional HTTP Basic Auth support for Tunarr web/admin routes.

When both `TUNARR_BASIC_AUTH_USER` and `TUNARR_BASIC_AUTH_PASSWORD` are configured, Tunarr will require Basic Auth for protected routes. If either env var is not set, Tunarr behaves the same as it does today.

The auth hook keeps media and integration endpoints public so external clients can continue working without credentials, including:

- HDHomeRun discovery / lineup endpoints
- XMLTV and M3U endpoints
- Stream endpoints
- Image and cached image endpoints
- Program artwork endpoints used by Jellyfin

## Why

This gives users a lightweight option for protecting Tunarr’s web/admin interface when running it on shared hosts, seedboxes, VPS environments, or other setups where binding to localhost or controlling the reverse proxy is not always practical.

## Testing

- Built successfully
- Deployed and tested on a seedbox
- Confirmed protected web/admin routes require Basic Auth when env vars are set
- Confirmed Tunarr still works when auth env vars are not set
- Confirmed stream/playback-related public endpoints continue to work
- Confirmed Jellyfin artwork loads without requiring auth


chrisbenincasa:
Thanks for the PR and sorry for the delay. I will take a look when I get a chance!

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+  reply: FastifyReply,
+) {
+  const expectedUser = process.env.TUNARR_BASIC_AUTH_USER;
+  const expectedPass = process.env.TUNARR_BASIC_AUTH_PASSWORD;
+
+  if (!expectedUser || !expectedPass) {
+    return;
+  }
+
+  if (isPublicPath(request.url)) {
```
chrisbenincasa: IMO this would be better implemented using each route's configuration object, rather than a regex (which is brittle)

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+    return false;
+  }
+}
+
+export async function tunarrBasicAuthHook(
+  request: FastifyRequest,
+  reply: FastifyReply,
+) {
+  const expectedUser = process.env.TUNARR_BASIC_AUTH_USER;
+  const expectedPass = process.env.TUNARR_BASIC_AUTH_PASSWORD;
```
chrisbenincasa: environment variables should live in `env.ts` which ensures proper type checking

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  if (!expectedUser || !expectedPass) {
```
chrisbenincasa: `isNonEmptyString`

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+function unauthorized(reply: FastifyReply) {
+  reply.header('WWW-Authenticate', 'Basic realm="Tunarr"');
+  return reply.status(401).send('Authentication required');
+}
+
+function safeEqual(a: string, b: string) {
+  const aBuf = Buffer.from(a);
+  const bBuf = Buffer.from(b);
+
+  if (aBuf.length !== bBuf.length) {
```
chrisbenincasa: I believe this is an attack vector in and of itself - attacker can enumerate usernames / passwords to find the right length (via timing attack w/ the early return). I believe the way around this is to always perform the safe equal on hashed versions of the inputs (always the same length)

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+  '/api/xmltv.xml',
+  '/api/channels.m3u',
+  '/favicon.ico',
+  '/favicon.svg',
+]);
+
+const PUBLIC_PREFIXES = [
+  '/stream/',
+  '/images/',
+  '/images/uploads/',
```
chrisbenincasa: this is redundant since we're already matching `/images` above. In any case I think the config-based approach per-route is more robust.

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+  '/lineup.json',
+  '/lineup_status.json',
+  '/api/xmltv.xml',
+  '/api/channels.m3u',
+  '/favicon.ico',
+  '/favicon.svg',
+]);
+
+const PUBLIC_PREFIXES = [
+  '/stream/',
```
chrisbenincasa: we will need the /streams/ prefix also to allow fetching of stream chunks

chrisbenincasa: also we should add just `/stream$` here - again not necessary to do all of these if we take the config-based approach

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+  } catch {
+    return false;
+  }
+}
+
+export async function tunarrBasicAuthHook(
+  request: FastifyRequest,
+  reply: FastifyReply,
+) {
+  const expectedUser = process.env.TUNARR_BASIC_AUTH_USER;
```
chrisbenincasa: this reads the environment on every request. it would be more simple to read the env on startup and conditoinally attach the hook if both fields are present.

chrisbenincasa (changes requested):
Left some comments. I think this also requires some automated (unit) tests.

Trifocals3537:
Thanks for the review, and no worries on the delay.

Your feedback makes sense. I built the first pass around a shared-host/seedbox use case where I couldn’t always rely on localhost binding or reverse proxy control, so I went with the quickest path that protected the UI without breaking Plex/Jellyfin/HDHR.

That said, I agree the path allowlist is too fragile for upstream. It’s too easy to miss a stream, artwork, guide, or client integration endpoint and create a weird playback/client issue.

I’ll rework this around route config instead, move the env vars into `env.ts`, read the config once at startup, fix the timing-safe comparison, and add tests.

Appreciate the guidance. I’ll push a cleaner version.

server/src/util/basicAuth.ts
Trifocals3537: Yeah, I agree. The allowlist was the quickest way to prove out the idea, but I can see how it gets fragile fast. I’ll move this to route config so public routes opt out intentionally instead of relying on path matching.

server/src/util/basicAuth.ts
Trifocals3537: Got it. I’ll move these into `env.ts` and use the existing env helper pattern.

server/src/util/basicAuth.ts
Trifocals3537: Yep, good call. I missed that helper.

server/src/util/basicAuth.ts
Trifocals3537: Good catch. I’ll fix this by hashing both values first and then comparing the fixed-length hashes with `timingSafeEqual`.

server/src/util/basicAuth.ts
Trifocals3537: Makes sense. This should go away once I remove the allowlist, but I’ll clean it up if any path fallback remains.

server/src/util/basicAuth.ts
Trifocals3537: Agreed. This is a good example of why the allowlist approach is fragile. I’ll move this to route config so stream chunk routes can be marked public directly.

Trifocals3537: Makes sense. Rather than keep adding one-off path cases here, I’ll remove the path allowlist and handle this through route config as suggested.

server/src/util/basicAuth.ts
Trifocals3537: Agreed. I’ll move the env lookup to startup and only register the hook when both auth values are present. That should keep the no-auth path untouched and avoid doing this work on every request.

Trifocals3537:
Thanks again. I pushed the follow-up changes.

I ended up reworking the implementation the way you suggested: route config now controls which routes opt out of auth, and the auth hook only gets registered when both auth env vars are set.

I also moved the env vars into `env.ts`, updated the comparison logic to hash before calling `timingSafeEqual`, and added focused unit coverage for the auth behavior.

Locally, the focused auth test passes:

`pnpm exec vitest --typecheck.tsconfig tsconfig.test.json --run src/util/basicAuth.test.ts`

And focused lint on the auth file passes:

`pnpm exec eslint src/util/basicAuth.ts --report-unused-disable-directives --max-warnings 0`

Small note: after rebasing onto latest `main`, full server typecheck is now going through the new `tsgo` setup and I’m seeing unrelated errors outside this PR. I left those alone and kept this update scoped to the auth changes.

chrisbenincasa:
Thanks for applying the suggestions! `main` builds clean for me locally. Let me checkout this branch and see what's up.

chrisbenincasa:
Your branch builds cleanly for me locally too, so it might be something on your end?

server/src/util/basicAuth.ts
```diff
@@ -0,0 +1,80 @@
...
+    const idx = decoded.indexOf(':');
+
+    if (idx < 0) {
+      return false;
+    }
+
+    const user = decoded.slice(0, idx);
+    const pass = decoded.slice(idx + 1);
+
+    return safeEqual(user, expectedUser) && safeEqual(pass, expectedPass);
```
chrisbenincasa: note - this is another vector for timing attack. the short-circuit of the user comparison means that the password would not be checked, allowing a potential attacker to determine which of the 2 components was wrong. you want to perform both comparisons unconditionally and then return the and'd result.

chrisbenincasa (changes requested):
Thanks again - one note about another potential timing attack vector. Besides that, I think this is looking pretty good.

chrisbenincasa:
~~Also sorry, I changed the base branch here to be `dev` but now this requires another rebase to remove commits that I've since merged into dev. It should rebase cleanly (I tested it on my end), but I don't think i have push permission on this PR or fork to do it myself.~~ nvm I got it.

server/src/util/basicAuth.ts
Trifocals3537: Good catch. I pushed a small fix so both comparisons run before returning the combined result.

chrisbenincasa (approved):

PR #1874, opened 2026-05-15
feat: integrate new stream selector with the ffmpeg pipeline

### LINKED ISSUE #1230: More Granular Audio/Subtitle Configuration
I assume per-channel audio is in the plans already the language the per-channel section for the subtitles say 'Override global audio and subtitle settings for this channel', but if it was possible it'd be great to pick at the 'Import Media' phase.

I would imagine you could add a dropdown control, somewhat similar to how TV shows expand to let you pick what seasons you'd like to add, or some other kind of modal that lets you either pick from a list or enter the track or subtitle name specifically. 

For example, if you wanted to include a Hong Kong movie with its original audio language and English subtitles, but have it on the same channel as an American movie without subtitles. Or be able to include a channel of movies that just have commentary tracks, such as a RiffTrax muxed into the video file.

Or if you have an archival release of an anime that includes multiple versions of an English dub, the DVD version of the Japanese audio, or the remastered version of the original Japanese broadcast audio. Picking just "English" or "Japanese" doesn't guarantee you'd get the audio tracks desired. 

Understandable if the Plex/Emby/Jellyfin API doesn't expose this data enough to make it possible.

chrisbenincasa:
Wires the profile-based stream selector into the passthrough (remux / HlsDirectV2) subtitle path, which still used the legacy `SubtitleStreamPicker`. The transcode path already went through `StreamSelector` on `main`, so after this change both paths resolve subtitles the same way and channel/program/filler stream-selection profiles finally apply to passthrough output.

## What changed

- `FfmpegStreamFactory.buildPassthroughSubtitles` now calls `StreamSelector` instead of `SubtitleStreamPicker`, and the now-dead `IChannelDB` dependency is dropped from the factory.
- New `StreamSelector.selectSubtitleStream` / `evaluateSubtitleSelection` resolve subtitles without selecting audio, since passthrough copies the source audio untouched.
- `preferTextBased` added to the subtitle action schema, plus a runtime `hints` channel on the evaluator.
- Shared `findMatchingRule` and `sortSubtitleCandidates` helpers replace logic that was duplicated across the two evaluator entry points and the two subtitle action branches.

## Bug fixed on `main`

The legacy picker ordered candidates with `orderBy(subtitleStreams, isImageBasedSubtitle, 'desc')`, which puts `true` — image-based — *first*. So "prefer text-based" preferred image-based subs, which the factory then discarded, losing subtitles entirely even when a matching text sub existed. The new comparator (`aImage - bImage`) sorts correctly and is stable.

## Behavior changes worth review

**`preferTextBased` splits into preference vs. capability.** Skipping subtitle extraction is gated on the runtime hint only; a profile's `preferTextBased` field affects sort order alone. The hint means "this caller can mux an embedded stream straight from the container", which is true of passthrough and not of transcode. Without the split, a profile with `preferTextBased: true` could hand the transcode path an embedded sub with no extracted file path, and with `webvttSidecarEnabled` off the pipeline would emit `subtitles=<video file>` — ffmpeg told to load the video as a subtitle source.

**Untagged subtitle streams no longer match every language preference.** The legacy picker skipped the language check when a stream carried no ISO 639-2 tag, so untagged subs matched any preference by accident. `streamMatchesLanguage` now requires a real match. Users with by-language preferences who were getting subtitles from untagged streams will stop.

**Channel subtitle preferences collapse into one `by_language` action.** This already applied to the transcode path; passthrough previously walked each preference with its own `filterType`. Given `[eng/forced (prio 0), spa/any (prio 1)]`, the collapsed action requires *both* languages to be forced, where the legacy picker tried forced-eng and then any-spa. Edge case, but a real semantic change for passthrough.

**No-default fallback preserved.** The `default` subtitle action falls back to the first candidate when nothing carries `default: true`, matching the legacy `defaultStream ?? orderedStreams[0]`. External subs frequently have no default flag, and returning null there would silently drop them.

## Not included

The web API client is **not** regenerated for the new `preferTextBased` field. Doing so pulls in ~111 lines of pre-existing drift that breaks `@tunarr/web` typecheck — tracked separately in #2115. Nothing in the web UI consumes the field yet, so it is API-only surface for now.

## Testing

Server suite 1557 passing. New coverage: a direct `evaluateSubtitleSelection` suite (rule matching, no-match, disable, no-default fallback, and both sides of the hint/action split), the resolver's `subtitlesEnabled: false` branch, passthrough with `subtitlesEnabled` off, and passthrough on content with no audio streams.

Manual verification against real media is still advisable, since every factory test mocks `StreamSelector` and no automated test exercises a user-defined profile end to end.

### DIFF
--- a/server/src/ffmpeg/FFmpegModule.ts
+++ b/server/src/ffmpeg/FFmpegModule.ts
@@ -5,6 +5,7 @@ import { ContainerModule } from 'inversify';
 import type { ChannelOrm } from '../db/schema/Channel.ts';
 import { bindAssistedFactory } from '../util/assistedInject.ts';
 import { FfmpegInfo } from './ffmpegInfo.ts';
+import { StreamSelector } from './StreamSelector.ts';
 
 export type FFmpegAssistedFactory = (
   transcodeConfig: TranscodeConfigOrm,
@@ -18,7 +19,8 @@ const FFmpegModule = new ContainerModule(({ bind }) => {
     FfmpegStreamFactory,
   );
 
-  bind(FfmpegInfo).toSelf();
+  bind(FfmpegInfo).toSelf().inSingletonScope();
+  bind(StreamSelector).toSelf().inSingletonScope();
 });
 
 export { FFmpegModule };
--- a/server/src/ffmpeg/StreamSelectionEvaluator.ts
+++ b/server/src/ffmpeg/StreamSelectionEvaluator.ts
@@ -84,6 +84,10 @@ export function buildCelContext(
   };
 }
 
+export type StreamSelectionHints = {
+  preferTextBased?: boolean;
+};
+
 export async function evaluateStreamSelectionProfile(
   profile: StreamSelectionProfile,
   audioStreams: NonEmptyArray<AudioStreamDetails>,
@@ -91,6 +95,7 @@ export async function evaluateStreamSelectionProfile(
   celService: CelEvaluationService,
   celContext: StreamSelectionCelContext,
   lineupItem: ContentBackedStreamLineupItem,
+  hints?: StreamSelectionHints,
 ): Promise<StreamSelectionResult> {
   for (const rule of profile.rules) {
     const conditionResult = celService.evaluate(rule.condition, celContext);
@@ -105,6 +110,7 @@ export async function evaluateStreamSelectionProfile(
         rule.subtitleAction,
         subtitleStreams,
         lineupItem,
+        hints,
       );
       return { audioStream, subtitleStream };
     }
@@ -178,6 +184,7 @@ async function resolveSubtitleAction(
   action: SubtitleAction,
   subtitleStreams: SubtitleStreamDetails[] | undefined,
   lineupItem: ContentBackedStreamLineupItem,
+  hints?: StreamSelectionHints,
 ): Promise<SubtitleStreamDetails | null> {
   switch (action.type) {
     case 'disable':
@@ -187,10 +194,34 @@ async function resolveSubtitleAction(
       if (!subtitleStreams || subtitleStreams.length === 0) {
         return null;
       }
-      const defaultStream = subtitleStreams.find((s) => s.default);
+
+      const effectivePreferText =
+        hints?.preferTextBased || action.preferTextBased;
+
+      const candidates = [...subtitleStreams];
+      if (effectivePreferText) {
+        candidates.sort((a, b) => {
+          const aImage = isImageBasedSubtitle(a.codec) ? 1 : 0;
+          const bImage = isImageBasedSubtitle(b.codec) ? 1 : 0;
+          return aImage - bImage;
+        });
+      }
+
+      const defaultStream = candidates.find((s) => s.default);
       if (!defaultStream) {
         return null;
       }
+
+      // When preferTextBased is active and the default is an embedded text-based
+      // sub, return it directly — the caller will extract via the pipeline.
+      if (
+        effectivePreferText &&
+        defaultStream.type === 'embedded' &&
+        !isImageBasedSubtitle(defaultStream.codec)
+      ) {
+        return defaultStream;
+      }
+
       const extracted =
         await SubtitleStreamPicker.getSubtitleDetailsWithExtractedPath(
           lineupItem,
@@ -221,9 +252,21 @@ async function resolveSubtitleAction(
         return null;
       }
 
+      const effectivePreferText =
+        hints?.preferTextBased || action.preferTextBased;
+
+      const candidates = [...subtitleStreams];
+      if (effectivePreferText) {
+        candidates.sort((a, b) => {
+          const aImage = isImageBasedSubtitle(a.codec) ? 1 : 0;
+          const bImage = isImageBasedSubtitle(b.codec) ? 1 : 0;
+          return aImage - bImage;
+        });
+      }
+
       for (const lang of action.languages) {
         const langLower = lang.toLowerCase();
-        for (const stream of subtitleStreams) {
+        for (const stream of candidates) {
           // Language match
           if (
             stream.languageCodeISO6392?.toLowerCase() !== langLower &&
@@ -249,6 +292,16 @@ async function resolveSubtitleAction(
           // Image-based check
           if (!action.allowImageBased && isImageBasedSubtitle(stream.codec)) {
             continue;
+          }
+
+          // When preferTextBased is active and we have an embedded text-based
+          // sub, return it directly — the caller will extract via the pipeline.
+          if (
+            effectivePreferText &&
+            !isImageBasedSubtitle(stream.codec) &&
+            stream.type === 'embedded'
+          ) {
+            return stream;
           }
 
           // For embedded text-based subs, verify extraction
--- a/server/src/ffmpeg/StreamSelector.ts
+++ b/server/src/ffmpeg/StreamSelector.ts
@@ -13,12 +13,14 @@ import {
   buildCelContext,
   evaluateStreamSelectionProfile,
 } from './StreamSelectionEvaluator.ts';
+import type { StreamSelectionHints } from './StreamSelectionEvaluator.ts';
 
 type StreamSelectRequest = {
   channel: ChannelOrm;
   lineupItem: ContentBackedStreamLineupItem;
   audioStreams: NonEmptyArray<AudioStreamDetails>;
   subtitleStreams: Array<SubtitleStreamDetails>;
+  hints?: StreamSelectionHints;
 };
 
 @injectable()
@@ -34,6 +36,7 @@ export class StreamSelector {
     lineupItem,
     audioStreams,
     subtitleStreams,
+    hints,
   }: StreamSelectRequest) {
     const selectionCtx = {
       channelId: channel.uuid,
@@ -61,6 +64,7 @@ export class StreamSelector {
       this.celService,
       celContext,
       lineupItem,
+      hints,
     );
   }
 }
--- a/server/src/services/StreamSelectionProfileResolver.ts
+++ b/server/src/services/StreamSelectionProfileResolver.ts
@@ -168,29 +168,36 @@ export class StreamSelectionProfileResolver {
           }
         : { type: 'default' as const };
 
-    // Build subtitle action from channel subtitle preferences
-    const subtitlePrefs =
-      await this.channelDB.getChannelSubtitlePreferences(channelId);
+    // Check if the channel has subtitles enabled
+    const channel = await this.channelDB.getChannel(channelId);
 
     let subtitleAction: StreamSelectionRule['subtitleAction'];
-    // Filter out preferences with filterType 'none' — they mean
-    // "don't match subtitles for this language", matching the
-    // behavior of SubtitleStreamPicker.pickSubtitles.
-    const activeSubtitlePrefs = subtitlePrefs.filter(
-      (p) => p.filterType !== 'none',
-    );
-    if (activeSubtitlePrefs.length > 0) {
-      const sorted = orderBy(activeSubtitlePrefs, 'priority', 'asc');
-      const topPref = sorted[0]!;
-      subtitleAction = {
-        type: 'by_language' as const,
-        languages: sorted.map((p) => p.languageCode),
-        filterType: topPref.filterType ?? 'any',
-        allowImageBased: Boolean(topPref.allowImageBased ?? true),
-        allowExternal: Boolean(topPref.allowExternal ?? true),
-      };
+    if (!channel?.subtitlesEnabled) {
+      subtitleAction = { type: 'disable' as const };
     } else {
-      subtitleAction = { type: 'default' as const };
+      // Build subtitle action from channel subtitle preferences
+      const subtitlePrefs =
+        await this.channelDB.getChannelSubtitlePreferences(channelId);
+
+      // Filter out preferences with filterType 'none' — they mean
+      // "don't match subtitles for this language", matching the
+      // legacy SubtitleStreamPicker behavior.
+      const activeSubtitlePrefs = subtitlePrefs.filter(
+        (p) => p.filterType !== 'none',
+      );
+      if (activeSubtitlePrefs.length > 0) {
+        const sorted = orderBy(activeSubtitlePrefs, 'priority', 'asc');
+        const topPref = sorted[0]!;
+        subtitleAction = {
+          type: 'by_language' as const,
+          languages: sorted.map((p) => p.languageCode),
+          filterType: topPref.filterType ?? 'any',
+          allowImageBased: Boolean(topPref.allowImageBased ?? true),
+          allowExternal: Boolean(topPref.allowExternal ?? true),
+        };
+      } else {
+        subtitleAction = { type: 'default' as const };
+      }
     }
 
     rules.push({
--- a/types/src/schemas/streamSelectionSchema.ts
+++ b/types/src/schemas/streamSelectionSchema.ts
@@ -36,10 +36,12 @@ export const SubtitleActionByLanguageSchema = z.object({
   filterType: SubtitleFilterSchema.default('any'),
   allowImageBased: z.boolean().default(true),
   allowExternal: z.boolean().default(true),
+  preferTextBased: z.boolean().default(false),
 });
 
 export const SubtitleActionDefaultSchema = z.object({
   type: z.literal('default'),
+  preferTextBased: z.boolean().default(false),
 });
 
 export const SubtitleActionSchema = z.discriminatedUnion('type', [
--- a/server/src/ffmpeg/FfmpegStreamFactory.ts
+++ b/server/src/ffmpeg/FfmpegStreamFactory.ts
@@ -30,7 +30,6 @@ import { isUndefined } from 'lodash-es';
 import type { DeepReadonly } from 'ts-essentials';
 import { match, P } from 'ts-pattern';
 import type { ContentBackedStreamLineupItem } from '../db/derived_types/StreamLineup.ts';
-import type { IChannelDB } from '../db/interfaces/IChannelDB.ts';
 import { FeatureFlagService } from '../services/FeatureFlagService.ts';
 import { isImageBasedSubtitle } from '../stream/util.ts';
 import { KEYS } from '../types/inject.ts';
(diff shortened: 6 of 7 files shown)


chrisbenincasa (commented):
## Review summary

Reviewed against head `2f8a2298` (local checkout matches PR head exactly). Verification: affected server test files pass **105/105**, `tsgo` typecheck is clean, oxlint on all changed non-test files reports 0 warnings/errors, and DI/container changes were cross-checked (`StreamSelector`'s deps are already singletons; the explicit binding correctly overrides the container's `autobind: true` transient binding).

**Positives**
- Real bug fix on `main`: legacy `SubtitleStreamPicker` used `orderBy(..., 'desc')` on a boolean to "prefer text-based", but `desc` puts `true` (image-based) *first* (verified empirically). So passthrough could pick an image sub, which the factory then discards — losing subtitles entirely even when a matching text sub existed. The new comparator (`aImage - bImage`) is correct and stable-sort-safe.
- Excellent ISO 639-2 B/T language matching (`ger` ↔ `deu`), case-insensitivity, and private-use code fallback, with regression tests for #1960.
- Passthrough finally honors stream selection profiles (the feature goal), and removing the now-dead `ChannelDB` dependency from `FfmpegStreamFactory` is clean.
- Embedded text-sub extraction failure now falls through to the next candidate instead of stopping.
- The factory test for "no audio streams but subtitles exist" directly covers the motivating scenario.

**Off-diff nits** (on lines this PR doesn't touch, so no inline comments):
- `buildCelContext` uses `s.index ?? 0` for subtitle indices (`StreamSelectionEvaluator.ts:71`) — two streams with an undefined index both become `0` in the CEL context, which can make CEL conditions ambiguous.
- Pre-existing typo `selectDefautlAudioStream` (`StreamSelectionEvaluator.ts:226`) — trivial to fix while in this file.

**Verdict:** solid, well-tested change. Before merge I'd resolve the two 🔴 items (the `preferTextBased` bypass gating, and the no-default-stream passthrough parity) and add the missing direct tests. Posted as `COMMENT`; happy to flip to APPROVE or REQUEST_CHANGES.

server/src/ffmpeg/StreamSelectionEvaluator.ts
```diff
@@ -207,10 +247,34 @@ async function resolveSubtitleAction(
...
+      }
+
+      const defaultStream = candidates.find((s) => s.default);
       if (!defaultStream) {
         return null;
       }
+
+      // When preferTextBased is active and the default is an embedded text-based
+      // sub, return it directly — the caller will extract via the pipeline.
+      if (
```
chrisbenincasa, line 294: 🔴 **Major — the extraction bypass can reach the transcode path and produce an invalid ffmpeg filter.**

This early return hands back an embedded text sub without its extracted file path whenever `hints?.preferTextBased || action.preferTextBased` (the same shortcut exists in the `by_language` branch below). That's correct for the passthrough caller, which remuxes the embedded stream with `SubtitleMethods.Convert` and lets the pipeline extract it.

But the transcode caller (`buildTranscodeInputs`) uses this same evaluator with no hints. If a profile sets the new `preferTextBased: true` field (API-created profiles; the UI doesn't expose it yet), an embedded text sub comes back with no `path`, so `source` falls back to the *video* source, and with `webvttSidecarEnabled` off the pipeline emits `SubtitlesInputSource(..., Burn)` → `SubtitleFilter` → `subtitles=<video path>`, i.e. ffmpeg is told to load the video as a subtitle file. On `main` this was impossible: embedded text subs always went through `getSubtitleDetailsWithExtractedPath`, and extraction failure returned `null` (no subtitle input at all).

Suggestion: gate the extraction bypass on `hints.preferTextBased` only (caller capability), and treat `action.preferTextBased` as a sorting preference — or have the transcode path extract/reject when it needs a real path.

server/src/ffmpeg/StreamSelectionEvaluator.ts
```diff
@@ -207,10 +247,34 @@ async function resolveSubtitleAction(
...
+      const candidates = [...subtitleStreams];
+      if (effectivePreferText) {
+        candidates.sort((a, b) => {
+          const aImage = isImageBasedSubtitle(a.codec) ? 1 : 0;
+          const bImage = isImageBasedSubtitle(b.codec) ? 1 : 0;
+          return aImage - bImage;
+        });
+      }
+
+      const defaultStream = candidates.find((s) => s.default);
```
chrisbenincasa: 🔴 **Major — passthrough parity regression: subtitles silently dropped when no stream is default-flagged.**

The legacy `SubtitleStreamPicker.pickSubtitles` (used by the passthrough path on `main`) fell back to `defaultStream ?? orderedStreams[0]` when the channel had no subtitle preferences. This `'default'` action returns `null` when no stream has `default: true`. So a channel with subtitles enabled, no subtitle preferences, and content whose subs lack a default flag (common for external subs) previously got subtitles in passthrough mode and now gets none. This matches the pre-existing transcode-mode semantics, so the PR unifies to the stricter behavior — fine if intentional, but it's a user-visible change worth calling out. Suggestion: fall back to the first candidate here to restore parity.

server/src/ffmpeg/StreamSelectionEvaluator.ts
```diff
@@ -119,6 +125,39 @@ export async function evaluateStreamSelectionProfile(
   };
 }
 
+/**
+ * Resolve only the subtitle stream, for callers that have no audio to select.
+ * Passthrough output reuses the source audio as-is, so requiring an audio
+ * stream here would drop subtitles from content that has none.
+ */
+export async function evaluateSubtitleSelection(
```
chrisbenincasa, line 144: 🟡 **Minor — the new subtitle-only path has no direct test coverage.**

`evaluateSubtitleSelection` isn't imported in `StreamSelectionEvaluator.test.ts` (which otherwise covers the evaluator very well), and the factory tests mock `StreamSelector`, so `selectSubtitleStream` → this function is never exercised end-to-end. Given the parity subtleties around it, add direct unit tests mirroring the `evaluateStreamSelectionProfile` suite: rule matching, no-match fallback → `null`, and disable/default/by_language with `preferTextBased` and extraction fallthrough.

server/src/ffmpeg/StreamSelectionEvaluator.ts
```diff
@@ -119,6 +125,39 @@ export async function evaluateStreamSelectionProfile(
...
+  profile: StreamSelectionProfile,
+  subtitleStreams: SubtitleStreamDetails[] | undefined,
+  celService: CelEvaluationService,
+  celContext: StreamSelectionCelContext,
+  lineupItem: ContentBackedStreamLineupItem,
+  hints?: StreamSelectionHints,
+): Promise<SubtitleStreamDetails | null> {
+  for (const rule of profile.rules) {
```
chrisbenincasa: 💡 **Nit — duplicated rule-matching loop.** This duplicates the loop in `evaluateStreamSelectionProfile` (including the debug log). A shared `findMatchingRule(profile, celService, celContext)` helper would keep the two from drifting. The `preferTextBased` candidate-sort comparator in `resolveSubtitleAction` is likewise duplicated across the two action branches.

server/src/ffmpeg/StreamSelectionEvaluator.ts
```diff
@@ -207,10 +247,34 @@ async function resolveSubtitleAction(
       if (!subtitleStreams || subtitleStreams.length === 0) {
         return null;
       }
-      const defaultStream = subtitleStreams.find((s) => s.default);
+
+      const effectivePreferText =
+        hints?.preferTextBased || action.preferTextBased;
```
chrisbenincasa: 💡 **Nit — two small things around this flag.** (1) The `||` OR semantics mean a profile's `preferTextBased: true` can never be overridden to `false` at runtime (hints are OR-only) — fine for the current single caller, but worth noting if a caller ever needs to force burn-in. (2) With the early return above, the `defaultStream &&` guard at lines 287–289 is always dead code — `defaultStream` is already narrowed non-null by the `if (!defaultStream) return null;` at 264–266.

server/src/ffmpeg/StreamSelectionEvaluator.ts
```diff
@@ -241,8 +305,20 @@ async function resolveSubtitleAction(
...
+          const bImage = isImageBasedSubtitle(b.codec) ? 1 : 0;
+          return aImage - bImage;
+        });
+      }
+
       for (const lang of action.languages) {
-        for (const stream of subtitleStreams) {
+        for (const stream of candidates) {
           // Language match
           if (!streamMatchesLanguage(stream, lang)) {
```
chrisbenincasa, line 339: 🟡 **Minor — behavior change vs the legacy picker for untagged subtitle streams; confirm it's intended.**

Legacy `SubtitleStreamPicker` skipped the language check when a stream had no ISO 639-2 tag (`if (stream.languageCodeISO6392 && !codesMatch(...)) continue;`), so untagged subs matched *any* preference. `streamMatchesLanguage` now requires a real match (`codesMatch(undefined, x) === false`), which is almost certainly the correct behavior — but passthrough users who previously got subs from untagged streams (by accident) will lose them when they have by-language prefs. Suggest noting this in the PR description.

server/src/services/StreamSelectionProfileResolver.ts
```diff
@@ -168,29 +168,37 @@ export class StreamSelectionProfileResolver {
...
+      const activeSubtitlePrefs = subtitlePrefs.filter(
+        (p) => p.filterType !== 'none',
+      );
+      if (activeSubtitlePrefs.length > 0) {
+        const sorted = orderBy(activeSubtitlePrefs, 'priority', 'asc');
+        const topPref = sorted[0]!;
+        subtitleAction = {
+          type: 'by_language' as const,
+          languages: sorted.map((p) => p.languageCode),
+          filterType: topPref.filterType ?? 'any',
```
chrisbenincasa, line 194: 🟡 **Minor — collapsing channel prefs into a single `by_language` action changes passthrough semantics for mixed `filterType` preferences.**

This collapse predates the PR for the transcode path, but passthrough previously used `SubtitleStreamPicker` directly, which applied each pref's own `filterType` while walking priorities. Now, e.g., prefs `[eng/forced (prio 0), spa/any (prio 1)]` require *both* languages to be forced, where the legacy picker tried forced-eng first and then any-spa. Edge case, but a conscious decision is worth recording.

server/src/services/StreamSelectionProfileResolver.ts
```diff
@@ -168,29 +168,37 @@ export class StreamSelectionProfileResolver {
           }
         : { type: 'default' as const };
 
-    // Build subtitle action from channel subtitle preferences
-    const subtitlePrefs =
-      await this.channelDB.getChannelSubtitlePreferences(channelId);
+    // Check if the channel has subtitles enabled
+    const channel = await this.channelDB.getChannel(channelId);
```
chrisbenincasa, line 172: 🟡 **Minor — extra DB round-trip + missing test.** This adds a `getChannel` query to every legacy-profile resolution (once per passthrough/transcode session) on top of the existing preference query; volume is low (session build time), so it's acceptable, but consider loading the channel once. Also: the new `!channel?.subtitlesEnabled → disable` branch below has no test — `createResolver` in the test file now plumbs `subtitlesEnabled`, but no test exercises `false`.

server/src/ffmpeg/FFmpegModule.ts
```diff
@@ -18,7 +19,8 @@ const FFmpegModule = new ContainerModule(({ bind }) => {
     FfmpegStreamFactory,
   );
 
-  bind(FfmpegInfo).toSelf();
+  bind(FfmpegInfo).toSelf().inSingletonScope();
```
chrisbenincasa: 💡 **Nit — unrelated scope creep.** `FfmpegInfo` → `inSingletonScope()` is safe (its only cache is `static`; `ChildProcessHelper` is stateless), but it's an independent change that would be cleaner in its own commit.

types/src/schemas/streamSelectionSchema.ts
```diff
@@ -36,10 +36,12 @@ export const SubtitleActionByLanguageSchema = z.object({
   filterType: SubtitleFilterSchema.default('any'),
   allowImageBased: z.boolean().default(true),
   allowExternal: z.boolean().default(true),
+  preferTextBased: z.boolean().default(false),
```
chrisbenincasa, line 39: 💡 **Nit — generated client will be stale.** Additive and safe (default `false`), but the committed `web/src/generated` client won't include `preferTextBased` until `pnpm generate-client` runs (no CI job checks freshness). Also note nothing in the web UI exposes this field yet, so it's currently API-only surface — which is exactly what makes the transcode Burn-path finding in `StreamSelectionEvaluator.ts` latent rather than immediately reachable.

server/src/ffmpeg/StreamSelector.ts
```diff
@@ -1,25 +1,34 @@
...
 type StreamSelectRequest = {
   channel: ChannelOrm;
   lineupItem: ContentBackedStreamLineupItem;
   audioStreams: NonEmptyArray<AudioStreamDetails>;
   subtitleStreams: Array<SubtitleStreamDetails>;
+  hints?: StreamSelectionHints;
+};
+
+type SubtitleSelectRequest = Omit<StreamSelectRequest, 'audioStreams'> & {
+  audioStreams?: Array<AudioStreamDetails>;
```
chrisbenincasa: 💡 **Nit — style inconsistency.** `Array<AudioStreamDetails>` here vs the `SubtitleStreamDetails[]` style used elsewhere in this file (and `NonEmptyArray` above); pick one convention.

chrisbenincasa:
Manual test plan for this branch, also committed at `architecture/stream-selector-manual-test-plan.md`.


## Scope

The PR wires `StreamSelector` into the ffmpeg pipeline, fixes two regressions the
integration introduced, and fixes two bugs an external review found. Unit tests
cover the evaluator. These cases cover what unit tests cannot — real containers,
real track layouts, real ffmpeg output.

## What is being verified

| Fix | Behavior before | Behavior after |
|---|---|---|
| Parity fallback | `default` action returned `null` when no track carried the default flag | Falls back to the first candidate |
| `audioDetails` requirement | Subtitle selection was skipped when a file had no audio streams | Subtitles resolve independently of audio |
| `subtitlesEnabled` gate | Passthrough path ignored the channel toggle | Passthrough honors `channel.subtitlesEnabled` |
| `preferTextBased` ordering | Image-based subs sorted first, so text subs lost | Text-based subs sort first |

## Why these cases

Track layouts were surveyed with ffprobe across a real library of roughly 600
movies and 240 TV shows. Three findings drive the plan.

- **Roughly 75–80% of files with subtitles carry no default flag**, in both
  movies and TV. The parity fallback therefore affects most playback, not an
  edge case.
- **Mixed image + text containers appear in about 14% of movies and none of the
  TV.** The ordering fix is effectively movies-only.
- **Files with zero audio streams are rare but real**, and at least one pairs
  zero audio with an external sidecar subtitle. That combination is what the
  `audioDetails` fix exists for.

Sourcing files for each case below is a matter of running ffprobe over a library
and filtering on stream counts and disposition flags.

## Setup

Create two channels over the same content.

- **Channel A** — passthrough / HlsDirectV2. Exercises `buildPassthroughSubtitles`.
- **Channel B** — transcode. Exercises `SubtitleFilter` burn-in.

The fixes sit on opposite sides of the pipeline, so every case runs on both.

## Cases

### 1. No-default fallback

**Needs:** a file with one or more subtitle tracks, none flagged default.

Set the subtitle action to `default`. Subtitles must appear. Before the fix this
returned `null` and subtitles were silently dropped.

### 2. Text-based preference beats image-first ordering

**Needs:** a file with many image-based subtitle tracks and exactly one text
track, none flagged default. A ratio around 30:1 is the sharpest available test.

With `preferTextBased` on, the single text track must win. Before the fix the
sort placed image tracks first and one of those was selected.

### 3. Ordering, second opinion

**Needs:** two or three more mixed files at a lower ratio, around 7:1.

Same assertion as case 2, to confirm the result is not an artifact of one file's
internal track order.

### 4. Subtitles with no audio, external sidecar

**Needs:** a video file with zero audio streams, zero embedded subtitle streams,
and a sidecar subtitle file beside it.

The sidecar must load. Before the fix, absent audio meant subtitle selection
never ran at all.

### 5. `by_language` with real competition

**Needs:** 5 or more audio tracks and 30 or more text subtitle tracks.

Pick a language present in both track sets. Confirm the audio and subtitle
selections agree, and that `preferChannels` resolves to the intended mix.

### 6. `by_language` across both subtitle types

**Needs:** multiple audio tracks, a large mixed image + text subtitle set, and at
least one subtitle track with no language tag.

Exercises `allowImageBased` in both positions and confirms untagged tracks do not
break the language matcher.

### 7. Forced-subtitle filter

**Needs:** a file with exactly one forced track among many, none flagged default.
Episodic TV is a reliable source.

Set `filterType: forced`. Only the forced track may be selected.

### 8. Untagged language does not crash the matcher

**Needs:** a file whose only subtitle track carries no language tag.

Run a `by_language` action. Selection must fall through cleanly rather than throw.

## Also worth a look

`buildPassthroughSubtitles` lacks the `isExternal` guard the transcode path
carries at `FfmpegStreamFactory.ts:533`. It is unreachable today because the
passthrough path only ever receives sidecar subs, but the asymmetry is fragile.
Not changed in this PR.

server/src/services/StreamSelectionProfileResolver.ts
chrisbenincasa: Keeping the `getChannel` call as-is. It runs once per session build on the legacy-profile path only, so the cost is one extra query plus a `transcodeConfig` join at stream setup. Narrowing it properly means adding a single-column accessor to `IChannelDB`, the `ChannelReadOpsRepository` implementation, and the test mock — more surface than the read is worth. Revisit if legacy profile resolution ever moves onto a hot path.

types/src/schemas/streamSelectionSchema.ts
chrisbenincasa: Not regenerating the client in this PR. `pnpm generate-client` pulls in ~111 lines of pre-existing drift that breaks the `@tunarr/web` typecheck, which is tracked in #2115. The field is additive and defaults to `false`, and nothing in the web UI reads it yet, so the stale client costs nothing today. Regenerating lands with #2115.

PR #1881, opened 2026-05-19
Feature/linkdaydisplay

redaloe:
Displays the day of a slot during linking in a weekly schedule to disambiguate slots happening on different days but the same time


chrisbenincasa:
Thanks! I actually just pushed the same change :) 31bce9c6548bec3e1f09bb8c06224e826f32b7cb

PR #1897, opened 2026-05-27
feat(server): add program_extra table and scanner hooks (Phase 0)

coreyjv:
Introduce the vendor-agnostic foundation for persisting program extras from remote media sources.

- Add `program_extra` table with dual nullable parent FKs covering all four content levels (movie, episode, show, season)
- Add `ProgramExtraMinter` injectable class for minting extra + artwork DAOs from `MediaSourceExtra` input
- Add `ProgramExtraRepository` and `upsertProgramExtras()` on `IProgramDB` / `ProgramDB`
- Add `MediaSourceExtra` type to `Media.ts` using `RemoteSourceType`
- Add no-op `scanExtrasForMovie()` hook to `MediaSourceMovieLibraryScanner`, called for both unchanged and newly-upserted movies
- Add no-op `scanExtrasForShow/Season/Episode()` hooks to `MediaSourceTvShowLibraryScanner` with equivalent call-site coverage

### DIFF
--- a/server/src/db/converters/ProgramExtraMinter.ts
+++ b/server/src/db/converters/ProgramExtraMinter.ts
@@ -8,6 +8,7 @@ import type { MediaSourceOrm } from '../schema/MediaSource.ts';
 import type { MediaSourceLibrary } from '../schema/MediaSourceLibrary.ts';
 import type { NewProgramExtra } from '../schema/ProgramExtra.ts';
 import type { NewProgramExtraWithRelations } from '../schema/derivedTypes.ts';
+import { MarkRequired } from 'ts-essentials';
 
 @injectable()
 export class ProgramExtraMinter {
@@ -40,14 +41,18 @@ export class ProgramExtraMinter {
       updatedAt: new Date(now),
     };
 
+    const mintedArtwork = extra.artwork.filter((art) =>
+      art.path !== undefined,
+    ) as MarkRequired<MediaArtwork, 'path'>[];
+
     return {
       extra: newExtra,
-      artwork: extra.artwork.map((art) => this.mintArtwork(art, uuid, now)),
+      artwork: mintedArtwork.map((art) => this.mintArtwork(art, uuid, now)),
     };
   }
 
   private mintArtwork(
-    artwork: MediaArtwork,
+    artwork: MarkRequired<MediaArtwork, 'path'>,
     programExtraId: string,
     now: number,
   ): NewArtwork {
--- a/server/src/db/converters/ProgramExtraMinter.ts
+++ b/server/src/db/converters/ProgramExtraMinter.ts
@@ -1,0 +1,63 @@
+import type { MediaArtwork } from '@tunarr/types';
+import dayjs from 'dayjs';
+import { injectable } from 'inversify';
+import { v4 } from 'uuid';
+import type { MediaSourceExtra } from '../../types/Media.ts';
+import type { NewArtwork } from '../schema/Artwork.ts';
+import type { MediaSourceOrm } from '../schema/MediaSource.ts';
+import type { MediaSourceLibrary } from '../schema/MediaSourceLibrary.ts';
+import type { NewProgramExtra } from '../schema/ProgramExtra.ts';
+import type { NewProgramExtraWithRelations } from '../schema/derivedTypes.ts';
+
+@injectable()
+export class ProgramExtraMinter {
+  mintExtra(
+    mediaSource: MediaSourceOrm,
+    library: MediaSourceLibrary,
+    extra: MediaSourceExtra,
+    parentProgramUuid: string | null,
+    parentGroupingUuid: string | null,
+    now: number = +dayjs(),
+  ): NewProgramExtraWithRelations {
+    const uuid = v4();
+
+    const newExtra: NewProgramExtra = {
+      uuid,
+      parentProgramUuid,
+      parentGroupingUuid,
+      extraType: extra.extraType,
+      title: extra.title,
+      summary: extra.summary ?? null,
+      duration: extra.duration,
+      externalKey: extra.externalId,
+      sourceType: extra.sourceType,
+      mediaSourceId: mediaSource.uuid,
+      libraryId: library.uuid,
+      filePath: extra.filePath ?? null,
+      canonicalId: extra.canonicalId ?? null,
+      state: 'ok',
+      createdAt: new Date(now),
+      updatedAt: new Date(now),
+    };
+
+    return {
+      extra: newExtra,
+      artwork: extra.artwork.map((art) => this.mintArtwork(art, uuid, now)),
+    };
+  }
+
+  private mintArtwork(
+    artwork: MediaArtwork,
+    programExtraId: string,
+    now: number,
+  ): NewArtwork {
+    return {
+      uuid: v4(),
+      programExtraId,
+      artworkType: artwork.type,
+      createdAt: new Date(now),
+      updatedAt: new Date(now),
+      sourcePath: artwork.path!,
+    };
+  }
+}
--- a/server/src/db/program/ProgramExtraRepository.ts
+++ b/server/src/db/program/ProgramExtraRepository.ts
@@ -1,0 +1,67 @@
+import { KEYS } from '@/types/inject.js';
+import { sql } from 'drizzle-orm';
+import { inject, injectable } from 'inversify';
+import { chunk } from 'lodash-es';
+import type { NewArtwork } from '../schema/Artwork.ts';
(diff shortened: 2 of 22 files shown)


server/src/db/schema/ProgramExtra.ts
```diff
@@ -0,0 +1,95 @@
...
+    createdAt: integer({ mode: 'timestamp_ms' }),
+    updatedAt: integer({ mode: 'timestamp_ms' }),
+  },
+  (table) => [
+    uniqueIndex('unique_program_extra').on(
+      table.sourceType,
+      table.mediaSourceId,
+      table.externalKey,
+    ),
+    index('program_extra_program_idx').on(table.parentProgramUuid),
```
chrisbenincasa: I would add `extraType` to both indexes here

server/src/migration/db/sql/0045_yummy_lucky_pierre.sql
```diff
@@ -0,0 +1,30 @@
...
+	FOREIGN KEY (`media_source_id`) REFERENCES `media_source`(`uuid`) ON UPDATE no action ON DELETE cascade,
+	FOREIGN KEY (`library_id`) REFERENCES `media_source_library`(`uuid`) ON UPDATE no action ON DELETE cascade,
+	CONSTRAINT "one_parent_required" CHECK(("program_extra"."parent_program_uuid" IS NOT NULL AND "program_extra"."parent_grouping_uuid" IS NULL)
+          OR ("program_extra"."parent_program_uuid" IS NULL AND "program_extra"."parent_grouping_uuid" IS NOT NULL))
+);
+--> statement-breakpoint
+CREATE UNIQUE INDEX `unique_program_extra` ON `program_extra` (`source_type`,`media_source_id`,`external_key`);--> statement-breakpoint
+CREATE INDEX `program_extra_program_idx` ON `program_extra` (`parent_program_uuid`);--> statement-breakpoint
+CREATE INDEX `program_extra_grouping_idx` ON `program_extra` (`parent_grouping_uuid`);--> statement-breakpoint
+ALTER TABLE `artwork` ADD `program_extra_id` text REFERENCES program_extra(uuid);--> statement-breakpoint
```
chrisbenincasa, line 29: was this generated before the code added the "on delete cascade" part? because it is missing here

server/src/db/converters/ProgramExtraMinter.ts
```diff
@@ -0,0 +1,63 @@
...
+    programExtraId: string,
+    now: number,
+  ): NewArtwork {
+    return {
+      uuid: v4(),
+      programExtraId,
+      artworkType: artwork.type,
+      createdAt: new Date(now),
+      updatedAt: new Date(now),
+      sourcePath: artwork.path!,
```
chrisbenincasa, line 65: This will throw if `path` is undefined/null

server/src/services/scanner/MediaSourceTvShowLibraryScanner.ts
```diff
@@ -115,6 +115,12 @@ export abstract class MediaSourceTvShowLibraryScanner<
         continue;
       }
 
+      await this.scanExtrasForShow(
+        context,
+        showResult.get()!,
```
chrisbenincasa: `.get()!` is called a few times (and one existing one) - not a bad time to extract out to a const at this point

server/src/services/scanner/MediaSourceTvShowLibraryScanner.ts
```diff
@@ -302,6 +308,12 @@ export abstract class MediaSourceTvShowLibraryScanner<
           continue;
         }
 
+        await this.scanExtrasForSeason(
+          scanContext,
+          persistedSeason.get()!,
```
chrisbenincasa: ditto re: making this ia const now

server/src/db/program/ProgramExtraRepository.ts
```diff
@@ -0,0 +1,68 @@
...
+            state: sql`excluded.state`,
+            updatedAt: sql`excluded.updated_at`,
+          },
+        })
+        .returning({ uuid: ProgramExtra.uuid })
+        .all();
+
+      for (let i = 0; i < batch.length; i++) {
+        const persistedUuid = upserted[i]?.uuid;
+        if (persistedUuid !== undefined) {
```
chrisbenincasa: isNonEmptyString

chrisbenincasa: Also - would this ever realistically be undefined or empty?

server/src/db/program/ProgramExtraRepository.ts
```diff
@@ -0,0 +1,68 @@
...
+            filePath: sql`excluded.file_path`,
+            canonicalId: sql`excluded.canonical_id`,
+            state: sql`excluded.state`,
+            updatedAt: sql`excluded.updated_at`,
+          },
+        })
+        .returning({ uuid: ProgramExtra.uuid })
+        .all();
+
+      for (let i = 0; i < batch.length; i++) {
```
chrisbenincasa, line 53: I think this could be slightly cleaner, avoid the issue with array index access nullish stuff, and avoid object spreads with something like:

```typescript
zip(upserted, batch).forEach(([{uuid: persistedUuid}, extra]) => {
  for (const art of extra.artwork) {
    art.programExtraId = persistedUuid;
    allArtwork.push(art);
  }
})
```

This assumes that the length of upserted and batch are strictly the same, which I think we can assume.

server/src/db/program/ProgramExtraRepository.ts
```diff
@@ -0,0 +1,68 @@
...
+          const artwork = batch[i]!.artwork.map((art) => ({
+            ...art,
+            programExtraId: persistedUuid,
+          }));
+          allArtwork.push(...artwork);
+        }
+      }
+    }
+
+    if (allArtwork.length > 0) {
```
chrisbenincasa: [nit] `upsertArtwork` protects against empty so you don't really have to do it here

chrisbenincasa (changes requested):
Wondering if you've thought at all about how we'll represent these in the search index?

server/src/db/converters/ProgramExtraMinter.ts
coreyjv: How would you like to handle this? `MediaArtwork` has `path` as nullish and `NewArtwork` (via `Artwork`) has `sourcePath` as non-null. Should I set it to an empty string?

server/src/db/program/ProgramExtraRepository.ts
coreyjv: This didn't actually resolve the `undefined` issues so I put in `isNonEmptyString` for now.

coreyjv:
> Wondering if you've thought at all about how we'll represent these in the search index?

I have not to be perfectly honest.

chrisbenincasa:
re: search index - like other things I'm guessing these will be nested objects on the `programs` index. noting because some special considerations have to be taken when dealing with nested objects on search documents depending on how we want to search / filter against them within the app, see: https://www.meilisearch.com/docs/resources/internals/datatypes#nested-document-querying-and-subdocuments

server/src/db/program/ProgramExtraRepository.ts
chrisbenincasa: Ah hm, I didn't think that drizzle would return that as optional on the query (I figured the nullish was just coming from the chaining on the array)

server/src/db/converters/ProgramExtraMinter.ts
chrisbenincasa: Yea the API is a little weird - I forgot I did the same thing here in other minters. The type is overloaded in how "path" is used. On the way out of the API, path is only set if the artwork is at a remote URL. When scanning, path is always set for remote sources. Local scanners bypass this type entirely and just deal in DAOs (and they set the artwork after minting).

For now, just the be safe, we should probably start fixing these minters up by setting the input type here as `MarkRequired<MediaArtwork, 'path'>` and then filtering the input before calling the method.

coreyjv:
I believe I've addressed all the comments I can. I'm not certain if that melisearch limitation is a blocker and if you want me to make specific changes related to it. 

Let me know if you want me to flip this PR to ready for review. Thanks!

chrisbenincasa:
Yup, let's un-draft it!

chrisbenincasa:
Rebase + promote to ready for review?

coreyjv:
I was waiting for when I had time to start immediately working on the other phases. It just hasn’t come around yet. 

I’m happy to do it if you don’t mind a lag between this being merged and the next phase coming. Let me know.

chrisbenincasa:
No rush!

Issue #1673, opened 2026-02-26
[Bug]: Local Library thumbs get 4xx errors from xmltv url


warrentc3:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [Discord](https://discord.gg/scMJxnPXaB) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/chrisbenincasa/tunarr/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Jellyfin Server stable, unstable or master; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.

### Description of the bug

On Version: 1.1.17

Using local libraries, not linked to a media server. Episode thumbs get 4xx error.

Also, I don't use Jellyfin, so requiring me to tick a box to create an issue that states I'm using an up to date JF is kinda odd.

### Reproduction steps

XMLTV and viewing guide in UI, img src url references ../api/programs/95b33ea5-be16-485b-beeb-5b663c5368ce/thumb?

Returns an HTTP 405 error

howver, if I navigate to ../api/programs/95b33ea5-be16-485b-beeb-5b663c5368ce/artwork/thumbnail 
the image renders correctly.

### What is the current _bug_ behavior?

XMLTV and viewing guide in UI, img src url references ../api/programs/95b33ea5-be16-485b-beeb-5b663c5368ce/thumb?

Returns an HTTP 405 error

### What is the expected _correct_ behavior?

a working URL that returns the episode image

### Tunarr version

Latest

### FFMPEG encoder type

None

### Deployment Type

Docker

### Full server logs

```shell
will attach if actually necessary.  should be an easy repro
```

chrisbenincasa:
This only happens when there is not a poster image for your item, but there is a thumbnail image saved in Tunarr. We can probably fallback to a thumb if poster isn't persisted or present on a scan. Simple workaround for you is to ensure you have poster images for your items. I don't think episodes support that _technically_ (https://kodi.wiki/view/Artwork/Episode) so I'll work on a fix there

warrentc3:
[...]

PR #1913, opened 2026-06-11
fix(api): fix /thumbs endpoint to properly return thumbnails, or posters as a fallback, for local programs

RichardMartija:


RichardMartija:
I put the wrong issue number in when I made my branch but I fixed the commit message at least. This is meant to address #1673

server/src/api/programmingApi.ts
```diff
@@ -855,6 +855,30 @@ export const programmingApi: RouterPluginAsyncCallback = async (fastify) => {
...
+              return res.status(404).send();
+            }
+            const path = req.serverCtx.imageCache.getImagePath(
+              art.cachePath,
+              art.artworkType,
+            );
+            return res.sendFile(
+              trimStart(
+                path.replace(globalOptions().databaseDirectory, ''),
+                '/',
```
chrisbenincasa: should we use the OS specific path sep? i think this is exposed as a constant in node; current impl wouldnt work for Windows?

RichardMartija: I copied this line from `/artwork` so that endpoint is gonna have the same problem if it exists.

RichardMartija: `fastify-static` seems to be calling `path.join` on the string here which normalizes the path so I think we might not even need to remove the leading slash at all

RichardMartija: Confirmed, it's not actually needed so I took it out all together.

server/src/api/programmingApi.ts
```diff
@@ -855,6 +855,27 @@ export const programmingApi: RouterPluginAsyncCallback = async (fastify) => {
...
+            }
+            if (!art || !art.cachePath) {
+              return res.status(404).send();
+            }
+            const path = req.serverCtx.imageCache.getImagePath(
+              art.cachePath,
+              art.artworkType,
+            );
+            return res.sendFile(
+              path.replace(globalOptions().databaseDirectory, ''),
```
chrisbenincasa, line 872: I think you will need a trimStart(x, '/') here

RichardMartija: I didn't encounter an issues without it in a linux environment. My understanding is that this fragment is passed into `path.join()` later and `path.join(basepath, "foo")` should result in the same output as `path.join(basepath, "/foo")` or even `path.join(basepath, "\\foo")` due to path normalization.

chrisbenincasa: Ok, if you've tested sufficiently then I believe you. The trim is done elsewhere but it's certainly possible it is not necessary

chrisbenincasa:
I was working in a larger consolation and cleanup of these endpoints - can you try https://github.com/chrisbenincasa/tunarr/pull/1923 and see if it fixes your issue?

RichardMartija:
The refactor looks like it's working properly.
I went ahead rebased my band-aid to main in case you wanted to do more work on the refactor but if you think it's ready to go soon then go ahead and drop my fix.

chrisbenincasa:
Ack - merged the refactor.

PR #1917, opened 2026-06-12
fix(web): refresh midroll config panel when the breakRule type changes

RichardMartija:
Before this change, the midroll config tab would not update to use the new break rule until you navigated to the filler tab and back. 

https://github.com/user-attachments/assets/ba3d27db-6a2a-4d93-b6f6-e2bb0a3b506f

This change allows the form to update immediately when the dropdown changes.

https://github.com/user-attachments/assets/99e18761-cf1b-4b7f-94ff-1ae8b95de3f6


chrisbenincasa (approved):

chrisbenincasa:
We should rebase this on main as well.

Fixes generally go to main (stable)

Features go to dev, unless they are fixes for unreleased features or fixes that require breaking DB changes.

RichardMartija:
Got it, I'll rebase. I was going off the contributing page which says "Target the dev branch for all PRs".

chrisbenincasa:
Yep, I'll get that updated, it's outdated.

PR #1918, opened 2026-06-12
fix(server): make the include season filter properly handle season 0

RichardMartija:
I realized after the fact that the include seasons filter has the exact same logic bug I fixed in the exclude filter with #1915.


chrisbenincasa:
Can you change the base branch to main?

RichardMartija:
@chrisbenincasa Working on it but FYI [this file](https://github.com/chrisbenincasa/tunarr/blob/dev/docs/generated/tunarr-v1.2.0-dev.1-openapi.json) was checked in with conflict markers in dev so resolving the rebase is a little tricky.

chrisbenincasa:
That file is generated, you can just regen with "pnpm turbo generate-openapi" and check in what it generates

PR #1926, opened 2026-06-18
fix(web): fix pagination on the task setting page so it won't cut off tasks after first ten rows

RichardMartija:
For some reason, the tasks table was paginating without showing pagination controls which resulted in whichever two tasks were last in the list being inaccessible unless you sorted the table by a different column (in which case two other tasks would be cut off).


chrisbenincasa:
The issue here is the page is using the wrong component to mount the table. It's using MRT_Table, which is a building block component that doesnt include header/footer, etc. Switching from MRT_Table to `MaterialReactTable` will fix this without disabling pagination.

RichardMartija:
Thanks, I don't really know React well. PR updated to fix pagination instead of removing.

chrisbenincasa:
Thanks for the fix!

PR #1943, opened 2026-06-30
feat: add support for transcoding to E-AC3 and Opus audio formats

RichardMartija:


chrisbenincasa (changes requested):
I think we need to do the following:

1. use /ffmpeg-info on the client side to get the list of available audio encoders
2. grey out the Opus option is the encoder is not available on the given build of ffmpeg
3. [optional] add a tooltip explaining that the provided ffmpeg build does not have Opus available

We can then use this framework as a basis for adding more options to the app

RichardMartija:
If this method for disabling unsupported encoders seems satisfactory, I'm happy to submit a followup to use the same method for the video encoding dropdown.

chrisbenincasa:
Sorry - have been away for a bit and still catching up.

server/src/db/schema/TranscodeConfig.ts
chrisbenincasa, line 199: sorry to keep giving you the run around, but i wonder if this is a good time to remove this constraint since it makes adding new encoders difficult and require a full DB migration, which isn't ideal. it does make things a bit less 'safe', but as long as we have validation at the API level we should be OK. anybody who would go and mess with the DB directly could break things, but that's not a use-case we need to support. 

we could probably remove all of these checks tbh.

RichardMartija:
Removed the constraint but also since reading the HLS spec, I discovered that Opus is not officially supported so I tagged that codec as experimental in the docs and added E-AC3 which is the only offically supported codec which was not supported by Tunarr.

chrisbenincasa:
fwiw I think there are still quite a few codecs supported by the HLS spec that are not supported by Tunarr - https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Audio

docs/configure/ffmpeg/transcode_config.md
```diff
@@ -127,6 +127,8 @@ Controls the codec used to encode the output audio stream.
 
 - **AAC** — The default and most compatible format. Widely supported by all clients and streaming devices. Recommended for most setups.
 - **AC3 (Dolby Digital)** — Good choice for surround sound (5.1) content. Supported by most TVs and AV receivers, but not all streaming clients.
+- **E-AC3 (Dolby Digital Plus)** — Supports higher bitrates and more channels than AC3 but is less supported. Use if you want to stream 7.1 (8 channel) surround sound and have compatible clients/hardware.
+- **Opus** — Highly efficient compression with inexpensive transcode perfomance and better multi-channel support than AAC or MP3. Would be ideal for surround sound content except that TVs and AV recievers don't support it but still useful for clients which lack support for AC3 such as browser-based clients running on Linux due to nearly-universal software support.
```
chrisbenincasa: you mentioned in the comment that the docs would say this is not in spec / experimental but I don't see it here

web/src/components/settings/ffmpeg/TranscodeConfigAudioSettingsForm.tsx
```diff
@@ -35,6 +37,14 @@ const AudioFormats: DropdownOption<SupportedTranscodeAudioOutputFormats>[] = [
     description: 'MP3',
     value: 'mp3',
   },
+  {
+    description: 'E-AC3',
+    value: 'eac3',
+  },
+  {
+    description: 'Opus',
```
chrisbenincasa: maybe we should put `(experimental)` here too?

chrisbenincasa (approved):

chrisbenincasa:
I'll merge once the final comments are addressed!

RichardMartija:
> fwiw I think there are still quite a few codecs supported by the HLS spec that are not supported by Tunarr - https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Audio

I was going off of their [RFC](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-3.1.3) which doesn't mention any of those other codecs, it looks like this is the discrepancy:
> 2.25. The container format for xHE-AAC, Apple Lossless, FLAC, and APAC audio MUST be fMP4.

So it seems that they don't support those codecs as packed audio but they do support them inside an fmp4 container which isn't what we're doing here.

docs/configure/ffmpeg/transcode_config.md
RichardMartija: I must have messed up my merge, fixed.

chrisbenincasa:
Got it. Yes ffmpeg as a consumer had a lot of issues dealing with fmp4 when I attempted a switch a year ago. Unclear whether it's gotten better.

PR #1958, opened 2026-07-16
fix(ffmpeg): "Could not open encoder before EOF" (-22) on stream restart when watermark is enabled

nixsupernova-ux:
### Summary
This PR fixes an intermittent streaming crash (`error code: -22` / `Could not open encoder before EOF`) that occurs during mid-stream restarts when channel watermarks are enabled. It addresses the underlying issue originally reported in #1903 (which is currently closed to comments).

### Root Cause Analysis
The crash is not specific to hardware acceleration (VAAPI), source codecs, or midroll fillers. Instead, it stems directly from the watermark filter chain:
1. `WatermarkScaleFilter.ts` outputs `scale=${width}:-1`. The `-1` flag rounds the auto-calculated aspect-ratio height to the nearest integer, which can be an odd number.
2. `WatermarkPixelFormatFilter` forces the watermark into a 4:2:0 chroma subsampling layout (`yuva420p`) when transparency/opacity is active (e.g., `aa=0.26`).
3. 4:2:0 pixel formats strictly require both width and height to be even numbers. If the watermark's aspect ratio yields an odd height, the filter graph presents an invalid frame size downstream, causing the encoder initialization to fail with an invalid argument error (`-22`).

### Solution
Updated `WatermarkScaleFilter.ts` to use `-2` instead of `-1`. This forces FFmpeg to round the auto-calculated dimension to the nearest even number, matching the safe math patterns already utilized by Tunarr's main video `ScaleFilter.ts`.

### Verification & Testing
- Added a dedicated `vitest` unit test to ensure `-2` formatting behavior is retained.
- Real-world verification: Disabling watermarks completely resolved the crashes over an 11-hour playback test (where crashes previously occurred every 5-30 minutes). Applying this scaling logic ensures safe handling when watermarks are re-enabled.

### DIFF
--- a/server/src/ffmpeg/builder/filter/watermark/WatermarkScaleFilter.test.ts
+++ b/server/src/ffmpeg/builder/filter/watermark/WatermarkScaleFilter.test.ts
@@ -1,0 +1,13 @@
+import { describe, expect, test } from 'vitest';
+import { WatermarkScaleFilter } from './WatermarkScaleFilter';
+
+describe('WatermarkScaleFilter', () => {
+  test('should return scale filter with width and force even height calculation (-2)', () => {
+    const width = 192;
+    const filter = new WatermarkScaleFilter(width);
+
+    const result = filter.build();
+
+    expect(result).toBe(`scale=${width}:-2`);
+  });
+});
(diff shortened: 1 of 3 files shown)


chrisbenincasa:
nice, thanks for the fix!

nixsupernova-ux:
Oh no it failed! I'll look into it further. My apologies

chrisbenincasa:
No worries, that’s why the checks are there

nixsupernova-ux:
Hey! So honestly I was just vibe coding to make the program work for my purposes. I started coding 3 days ago lol but I really like this application and am happy to work on it further. I have some bugs I'm working on now, but I was curious if I went about doing things the right way? The guidelines/code of conduct seemed to be blank so I just want to make sure I'm more helpful than annoying. Keep up the great work! Thank you for caring about Tunarr

chrisbenincasa:
For sure. Be careful with vibing :) 

Contribution guidelines are on the website https://tunarr.com/dev/contributing/

PR #1963, opened 2026-07-22
fix(plex): reconcile program identity by plex-guid on library scan

### LINKED ISSUE #1962: Plex ratingKey identity desync: channels fail to play after reindex / library changes (architectural review)
**Audience:** [Tunarr](https://github.com/chrisbenincasa/tunarr) maintainers  
**Date:** 2026-07-21  
**Reporter:** Production operator (sanitized report — no library names, paths, or media titles)  
**Tunarr version observed:** 1.2.8  
**Code reviewed:** `main` @ [`015994f`](https://github.com/chrisbenincasa/tunarr/commit/015994fc83e86bbeb0572d8cb6b5ef64f1db60f0) (2026-07-21)

---

## Executive summary

A multi-channel Tunarr deployment backed by Plex experienced **channels that appeared to load forever and would not play**. Health checks stayed green. Logs showed repeated Plex **`404 Not Found`** responses for `/library/metadata/{ratingKey}`, followed by transcode failures and missing HLS playlists.

The failure was **not** simply “the library never scanned.” Scans ran, but **identity reconciliation between Plex and Tunarr is architecturally fragile**. When Plex items are moved, removed, reindexed, or when on-disk paths diverge from Plex metadata, Tunarr can retain **stale Plex rating keys** on the program rows that channels actually reference. Playback then fails even when the underlying media file still exists.

This report documents observed symptoms, maps them to **root causes in current Tunarr architecture**, cites relevant source locations, and proposes concrete fixes.

---

## 1. Observed symptoms

| Symptom | Detail |
|---------|--------|
| Client behavior | Channel tune hangs indefinitely (“loading”) |
| HLS API | `GET /stream/channels/{id}.m3u8?mode=hls` → HTTP **500**, “Stream not ready”, or missing `stream.m3u8` |
| Server logs | `Not found: …/library/metadata/{ratingKey}` → `Error while transcoding program stream` |
| Health endpoint | `/api/system/health` reported **healthy** while streams were failing |
| Operator assumption | “Plex didn’t finish scanning into Tunarr” — **misleading**: scans completed or appeared recent, but **identity and lineup bindings were stale** |

### Typical log signatures

```text
Not found: https://…/library/metadata/{ratingKey}
Error while transcoding program stream
Stream not ready yet. Retry
No playlist found … stream.m3u8
Failed to convert Plex API item … has invalid duration
Unable to parse schema from response … Stream[0].width / height expected number, received undefined
SqliteError: ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
msg: Error scanning library {libraryId}
```

---

## 2. When this breaks (trigger scenarios)

These operator actions are normal Plex maintenance. Tunarr should tolerate them; today it often does not.

| Trigger | What Plex does | What Tunarr does today | User-visible result |
|---------|----------------|------------------------|---------------------|
| **Library reindex / rebuild** | Numeric `ratingKey` values change for the same logical item | Creates **new** `program` rows keyed by new rating keys; **channels keep old program UUIDs** | 404 on old keys → playback failure |
| **Item removed from Plex** | Metadata endpoint 404 | Old program may remain in channel lineups | Infinite loading on that slot |
| **File moved/renamed on disk** | Plex `Part.file` updates; rating key may or may not change | Scan may skip update when `canonicalId` unchanged; paths in DB drift | Direct file fallback fails; network path 404 |
| **Scan partial failure** | N/A | `lastScannedAt` still advances; errors only logged | UI shows “recent scan” while DB is inconsistent |
| **Plex metadata quirks** | Missing stream width/height or zero duration | Hard-fail item conversion | Items never upserted; lineups reference ghosts |
| **Concurrent channel tunes** | N/A | Transcode session / RAM disk pressure | Same HTTP 500 symptoms as Plex failures (misleading) |

---

### 3.1 Plex `ratingKey` is treated as the durable program identity

**Problem:** Plex `ratingKey` is a **server-local, mutable** identifier. It changes on library reindex, section moves, and some repair operations. Tunarr stores it as `program.externalKey` and uses it as the upsert conflict key.

**Schema — unique constraint on rating key:**

```111:118:server/src/db/schema/Program.ts
    uniqueIndex(
      'program_source_type_external_source_id_external_key_unique',
    ).on(table.sourceType, table.externalSourceId, table.externalKey),
    uniqueIndex('program_source_type_media_source_external_key_unique').on(
      table.sourceType,
      table.mediaSourceId,
      table.externalKey,
    ),
```

**Upsert — conflict target is `(sourceType, mediaSourceId, externalKey)`:**

```87:96:server/src/db/program/ProgramUpsertRepository.ts
            .insert(Program)
            .values(c.map(({ program }) => program))
            .onConflictDoUpdate({
              target: [
                Program.sourceType,
                Program.mediaSourceId,
                Program.externalKey,
              ],
              set: ProgramUpsertSetClause,
            })
```

**Minting — `externalKey` is set from Plex `externalId` (the rating key):**

```104:108:server/src/db/converters/ProgramMinter.ts
    const newMovie = {
      uuid: programId,
      sourceType: movie.sourceType,
      externalKey: movie.externalId,
```

**Effect after Plex reindex:**

1. Plex lists the item under a **new** rating key.
2. Upsert cannot match the existing row (different `externalKey`) → **inserts a new `program` UUID**.
3. Channel lineups (`channel_programs` + persisted `channel-lineups/*.json`) still reference the **old UUID**.
4. Playback loads the old row → Plex GET by stale key → **404** → transcode abort.

**Fix needed:** Use stable identifiers (`plex-guid`, agent IDs, or file fingerprint) as the primary merge key. On match, **update `externalKey` in place** and migrate channel references — never fork a new program row for the same logical item.

---

### 3.2 `canonicalId` is content-hash based and intentionally ignores rating key / GUID changes

Tunarr computes `canonicalId` from Plex metadata hashes (title, parts, media IDs, etc.), **not** from `plex-guid` or `ratingKey`.

```42:62:server/src/services/PlexMediaCanonicalizers.ts
  private canonicalizePlexMovie(plexMovie: PlexMovie): string {
    const hash = crypto.createHash('sha1');
    hash.update(plexMovie.key);
    hash.update(plexMovie.title);
    // … part IDs, file paths, sizes …
    return hash.digest('base64');
  }
```

The test suite **explicitly requires** that `ratingKey` and `guid` changes do **not** change `canonicalId`:

```237:240:server/src/services/PlexMediaCanonicalizers.test.ts
  it('does not change id when ratingKey changes', () => {
    expect(canonicalizer.getCanonicalId(baseMovie)).toBe(
      canonicalizer.getCanonicalId({ ...baseMovie, ratingKey: '999' }),
```

**Scan skip when `canonicalId` matches — no path/key refresh:**

```209:220:server/src/services/scanner/MediaSourceMovieLibraryScanner.ts
    if (
      !force &&
      existingMovie &&
      existingMovie.canonicalId &&
      existingMovie.canonicalId === incomingMovie.canonicalId
    ) {
      this.logger.debug(
        'Found an unchanged program: rating key = %s, program ID = %s',
        fullMovie.externalId,
        existingMovie.uuid,
      );
      return;
    }
```

**Effect:**

- When Plex changes **only** the rating key (common after reindex) but file metadata hash is unchanged, a scan that *does* find the old row by key may **skip all updates**.
- When the rating key changes, lookup is by **new** key (`existingPrograms[incomingMovie.externalId]`) — old row is orphaned, not updated.

**Fix needed:**

- Lookup existing programs by **`plex-guid` first**, then canonicalId, then file path — never only by current rating key.
- On GUID match with new rating key: update `program.externalKey`, external ID rows, and media file paths atomically.
- Do not treat `canonicalId` equality as “unchanged” if `externalKey` differs.

---

### 3.3 Multiple identity stores desynchronize

The same Plex item is represented in several places that can disagree:

| Store | Role | Failure mode |
|-------|------|--------------|
| `program.externalKey` | Primary Plex rating key on program row | Used for API calls; stays stale |
| `program_external_id` (`source_type = 'plex'`) | Denormalized Plex key + paths | May be updated on scan while `program.externalKey` is not |
| `program_external_id` (`source_type = 'plex-guid'`) | Stable GUID | Not used as playback fallback |
| `program_media_file.path` / `direct_file_path` | On-disk path | Drifts when folders renamed |
| `channel_programs` + `channel-lineups/{id}.json` | Channel → program UUID binding | Points at superseded UUID after duplicate insert |

**External ID upsert does not update `externalKey` on conflict:**

```290:301:server/src/db/program/ProgramExternalIdRepository.ts
              .onConflictDoUpdate({
                target: [
                  ProgramExternalId.programUuid,
                  ProgramExternalId.sourceType,
                ],
                targetWhere: dbIsNull(ProgramExternalId.mediaSourceId),
                set: {
                  updatedAt: sql`excluded.updated_at`,
                  externalFilePath: sql`excluded.external_file_path`,
                  directFilePath: sql`excluded.direct_file_path`,
                  programUuid: sql`excluded.program_uuid`,
                },
```

(`externalKey` is absent from `set` for both single and multi-source conflict branches.)

**Fix needed:** Single transactional write path for all identity fields. Any scan or repair that learns a new rating key must update **every** denormalized copy and invalidate lineup caches if UUID changes.

---

### 3.4 Library scan errors: SQLite `ON CONFLICT` mismatch

**Observed in production:** repeated scan failures:

```text
SqliteError: ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
```

This matches Drizzle `onConflictDoUpdate` using `targetWhere` partial unique indexes on `program_external_id` and `program_grouping_external_id`. If migration-defined partial indexes do not exactly match the ORM conflict specification (or legacy DBs have drift), SQLite rejects the statement.

Relevant partial unique indexes:

```43:54:server/src/db/schema/ProgramExternalId.ts
    uniqueIndex('unique_program_multiple_external_id_media_source')
      .on(table.programUuid, table.sourceType, table.mediaSourceId)
      .where(sql`\`media_source_id\` is not null`),
    uniqueIndex('unique_program_single_external_id_media_source')
      .on(table.programUuid, table.sourceType)
      .where(sql`\`media_source_id\` is null`),
```

**Scan error handling — logged, not surfaced:**

```175:176:server/src/services/scanner/MediaSourceScanCoordinator.ts
        .catch((e) =>
          this.logger.error(e, 'Error scanning library %s', library.uuid),
```

**`lastScannedAt` advances even when per-item upserts fail silently:**

```127:137:server/src/services/scanner/MediaSourceScanner.ts
      await this.scanInternal({ … });
      await this.mediaSourceDB.setLibraryLastScannedTime(library.uuid, dayjs());
```

Individual movie upsert failures are caught and logged as warnings without failing the library scan:

```236:246:server/src/services/scanner/MediaSourceMovieLibraryScanner.ts
    const upsertResult = await Result.attemptAsync(() =>
      this.programDB.upsertPrograms([minted]),
    );
    if (upsertResult.isFailure()) {
      this.logger.warn(upsertResult.error, 'Error while processing movie (%O)', incomingMovie);
      return;
    }
```

**Fix needed:**

- Align every `ON CONFLICT` target with an exact SQLite unique index (add migration tests).
- Track per-library scan status: `success | partial | failed` with last error message.
- Do not update `lastScannedAt` when the scan transaction aborts or conflict rate exceeds threshold.
- Expose failed libraries in API/UI.

---

### 3.5 Strict Plex Zod parsing rejects valid-but-incomplete metadata

Video streams require `width` and `height` as numbers:

```278:279:types/src/plex/index.ts
  height: z.number(),
  width: z.number(),
```

Plex `Part.Stream` arrays are validated strictly:

```372:372:types/src/plex/index.ts
      Stream: z.array(PlexMediaStreamSchema).optional(),
```

**Hard-fail on invalid duration during movie injection:**

```1542:1547:server/src/external/plex/PlexApiClient.ts
    if (isNil(plexMovie.duration) || plexMovie.duration <= 0) {
      return Result.forError(
        new Error(
          `Plex movie ID = ${plexMovie.ratingKey} has invalid duration.`,
```

**Effect:** One bad Plex item can prevent upsert for that title. Combined with (3.4), operators cannot tell which lineup entries are unplayable until tune time.

**Fix needed:** Soft-fail per item → quarantine with reason, continue scan. Allow playback from `direct_file_path` / ffprobe when Plex stream dimensions are missing.

---

### 3.6 Playback path still depends on Plex network identity

**Metadata fetch always uses rating key:**

```704:705:server/src/external/plex/PlexApiClient.ts
    const responseResult = await this.doTypeCheckedGet(
      `/library/metadata/${key}`,
```

**Stream resolution** (`ProgramStreamDetailsFetcher`) tries local `mediaFiles` first, then falls back to Plex `externalFilePath` HTTP URL — not a true offline path:

```200:264:server/src/stream/ProgramStreamDetailsFetcher.ts
      const filePath = head(firstVersion.mediaFiles)?.path;
      const serverPath = program.externalIds.find(
        (eid) => eid.sourceType === server.type,
      )?.externalFilePath;
      // … fileExists → else network HttpStreamSource with Plex token URL
```

If `mediaFiles.path` is wrong/stale and `externalFilePath` embeds a dead rating key, playback fails.

**`plexStream.streamPath` (`network` | `direct`) appears unused in server streaming code.** The setting is defined and persisted:

```193:198:types/src/schemas/settingsSchemas.ts
export const PlexStreamSettingsSchema = z.object({
  streamPath: z.enum(['network', 'direct']).default('network'),
```

…but no server-side stream/FFmpeg code reads `settings.plexSettings().streamPath` (only the settings API and `SettingsDB` reference it). Operators enabling “direct” may believe they are bypassing Plex when the runtime does not implement that behavior.

**Fix needed:**

- Wire `streamPath` into `ProgramStreamDetailsFetcher` / FFmpeg input selection, **or** remove the setting from UI until implemented.
- On Plex 404: resolve current key via stored `plex-guid` → retry → persist.
- Prefer `direct_file_path` and local ffprobe over live Plex metadata for transcode decisions.

---

### 3.7 Channel lineup persistence decoupled from program identity repairs

Lineups are stored both relationally and as JSON files:

```217:217:server/src/db/channel/LineupRepository.ts
      `channel-lineups/${channelId}.json${isDelete ? '' : '.bak'}`,
```

When scans create duplicate program rows, **lineups are not automatically repointed** to the canonical row. Stale UUIDs persist until a manual lineup rebuild or external DB surgery.

`TvGuideService` acknowledges lineup vs DB duration drift but relies on scheduled tasks to heal:

```865:873:server/src/services/TvGuideService.ts
      if (currentProgram.startTimeMs < nextOffsetTime) {
        // … durations of the lineup and DB item can go out of sync …
        // We'll just push a pad and move on - a scheduled task will resolve
        // these discrepancies.
```

**Fix needed:** On program identity merge, cascade UUID updates to `channel_programs` and rewrite lineup JSON. Provide a “validate lineups against Plex” maintenance task.

---

### 3.8 Health checks do not cover Plex identity integrity

Registered checks (FFmpeg, transcode dir, missing associations, etc.):

```12:19:server/src/services/health_checks/HealthCheckModule.ts
  bind<HealthCheck>(KEYS.HealthCheck).to(FfmpegDebugLoggingHealthCheck);
  bind<HealthCheck>(KEYS.HealthCheck).to(BaseImageHealthCheck);
  bind<HealthCheck>(KEYS.HealthCheck).to(MissingProgramAssociationsHealthCheck);
  // … no Plex stale-key or scan-failure check
```

**Fix needed:** Add checks for:

- Last library scan failure / `ON CONFLICT` errors
- Active channel lineups referencing `program.state = 'missing'` or Plex 404 on `externalKey`
- Count of duplicate programs sharing the same `plex-guid`

---

### 3.9 Schedule / transcode issues amplify confusion

**Schedule wrapping** exists in `StreamProgramCalculator` (`elapsed % channelDuration`), so “stuck past end” may be a separate `channel.startTime` API state issue — but unplayable items in the current slot still spin until manually skipped.

**Transcode directory exhaustion** (small RAM disk, concurrent sessions) produces HTTP 500s indistinguishable from Plex 404 failures in clients.

**Fix needed:** Classify errors in API responses (Plex identity vs filesystem vs ffmpeg vs capacity). Enforce concurrent session limits with clear backoff.

---

## 4. Failure flow (conceptual)

```mermaid
flowchart TD
  A[Plex reindex / move / remove] --> B[Tunarr library scan]
  B --> C{Upsert by ratingKey}
  C -->|New key| D[New program UUID inserted]
  C -->|ON CONFLICT error| E[Scan abort for library]
  C -->|canonicalId match| F[Skip update — stale key/paths remain]
  D --> G[Old program row orphaned]
  G --> H[Channel lineups still reference old UUID]
  H --> I[Stream uses stale program.externalKey]
  I --> J[Plex GET /library/metadata/oldKey → 404]
  J --> K[HLS 500 / client loading forever]
  E --> L[lastScannedAt still updated]
  L --> M[Health check stays green]
```

---

## 5. Evidence summary (sanitized)

| Check | Result |
|-------|--------|
| Tunarr version | 1.2.8 |
| Channels affected | All tuned channels (several dozen) |
| Programs in database | ~23k |
| Health API during outage | Healthy |
| Plex metadata 404s | High volume for `/library/metadata/{id}` |
| `program` vs `program_external_id` | Frequent desync after rematch attempts |
| UNIQUE on `(source_type, media_source_id, external_key)` | Prevented in-place key update when duplicate row existed |
| Scan errors | `ON CONFLICT clause does not match…` on multiple libraries every cycle |
| Zod parse errors | Missing `Stream[0].width` / `height`; invalid duration |
| Post-repair verification | Sequential HLS probe: playback restored on nearly all channels |

---

## 6. Recommended engineering backlog

| Priority | Item | Primary files |
|----------|------|---------------|
| **P0** | Rebind by `plex-guid` on scan; update `externalKey` in place; migrate channel UUIDs | `MediaSourceMovieLibraryScanner.ts`, `ProgramUpsertRepository.ts`, `LineupRepository.ts` |
| **P0** | Fix SQLite `ON CONFLICT` / partial unique index alignment | `ProgramExternalIdRepository.ts`, `ProgramGroupingUpsertRepository.ts`, migrations |
| **P0** | Stream startup: GUID-based rating key refresh on 404; file-first playback | `PlexApiClient.ts`, `ProgramStreamDetailsFetcher.ts` |
| **P1** | Include `externalKey` in external-ID upsert `set`; unify identity writes | `ProgramExternalIdRepository.ts` |
| **P1** | Stop skipping scans when `canonicalId` matches but `externalKey` differs | `MediaSourceMovieLibraryScanner.ts` (+ TV/music scanners) |
| **P1** | Implement or remove `plexStream.streamPath` | `settingsSchemas.ts`, `ProgramStreamDetailsFetcher.ts` |
| **P1** | Soft-fail bad Plex metadata; quarantine table + UI | `PlexApiClient.ts`, `types/src/plex/index.ts` |
| **P2** | Health: stale keys in active lineups, scan failure state | `HealthCheckModule.ts`, `MediaSourceScanCoordinator.ts` |
| **P2** | Distinct error classes for Plex vs transcode vs disk | `ProgramStream.ts`, `streamApi.ts` |
| **P2** | Lineup auto-heal when program merged/superseded | `LineupRepository.ts`, `TvGuideService.ts` |

---

## 7. Suggested regression tests

1. **Rating key rotation:** Simulate Plex returning a new `ratingKey` for an unchanged `guid` → assert same `program.uuid`, updated `externalKey`, channels still play.
2. **Reindex orphan:** Old UUID in lineup → scan → automatic repoint or explicit repair task.
3. **ON CONFLICT:** Upsert external IDs on DB migrated from pre–`media_source_id` schema → no SQLite conflict error.
4. **Zod soft-fail:** Plex item with missing `width`/`height` → quarantined, scan continues.
5. **404 recovery:** Stream start with stale key but valid `plex-guid` and file path → self-heal and play.
6. **Health:** Inject stale lineup reference → health reports non-healthy with actionable message.

---

## 8. What is outside Tunarr (but exposes these bugs)

- Operator file renames / folder structure changes (Tunarr should heal via successful scan + path replacement).
- Media genuinely deleted from disk.
- Undersized transcode volume under heavy concurrent load.

These are manageable operationally **if** Tunarr fails loudly and self-heals identity after Plex changes.

---

## 9. Offer to maintainers

Happy to provide:

- Redacted log excerpts (no library names or paths)
- Sanitized SQLite schema statistics
- A minimal reproduction database for automated regression

Please treat **Plex `ratingKey` as ephemeral** and **`plex-guid` + durable file identity as canonical**. The current combination of rating-key upserts, duplicate program rows, scan conflict failures, unused “direct stream” settings, and green health checks produces exactly the “channels won’t play after Plex changed” failure mode operators report.

---

*Repository: [github.com/chrisbenincasa/tunarr](https://github.com/chrisbenincasa/tunarr)*

roto31:
## Summary

Fixes playback failures after Plex library reindex / rating key rotation ([#1962](https://github.com/chrisbenincasa/tunarr/issues/1962)).

When Plex assigns a new `ratingKey` to the same logical item, Tunarr previously inserted a **new** `program` row (upsert conflict key is `externalKey`) while channel lineups kept referencing the **old** program UUID. Streams then called `/library/metadata/{staleKey}` → 404 → HLS failure.

This PR (movie libraries first):

- **`PlexProgramIdentityService`** — resolve existing programs by `plex-guid` (skipping synthetic `local://` guids) and `canonicalId` when the current rating key is not found
- **Scan skip guard** — only skip when `canonicalId` matches **and** the rating key is unchanged
- **`ProgramDB.reconcilePlexRatingKeyChange`** — update `program.external_key` in place; refresh `program_external_id`; if a duplicate row already owns the new key, remap `channel_programs` + `channel-lineups/*.json` and mark the duplicate `missing`
- **`ProgramExternalIdRepository`** — include `externalKey` in `onConflictDoUpdate` set; add `lookupProgramSummaryByPlexGuid`
- **`PlexIdentityDesyncHealthCheck`** — warn when active lineups reference programs in `missing` state

## Test plan

- [x] `plexProgramIdentityUtil.test.ts` — guid extraction, `local://` detection
- [x] `PlexProgramIdentityService.test.ts` — rating key rotation, skip logic
- [x] `PlexIdentityDesyncHealthCheck.test.ts` — healthy vs warning paths
- [ ] `pnpm test` in `server/` (CI)
- [ ] Manual: Plex reindex → force library scan → channel playback without stale 404s

## Follow-up PRs (not in this change)

| Item | Notes |
|------|--------|
| TV / music / other scanners | Same identity path as movies |
| `plexStream.streamPath` | Wire `direct` in `ProgramStreamDetailsFetcher` (setting currently unused) |
| Stream-time 404 recovery | Resolve via `plex-guid` at playback |
| Zod soft-fail | Quarantine items with missing width/height or invalid duration |
| SQLite `ON CONFLICT` migration test | Partial unique index alignment for legacy DBs |

Closes #1962 for the movie-library identity path; issue can stay open for follow-ups if preferred.


scripts/notarize-macos-dmg.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
```
chrisbenincasa, line 1: what are all of these for? This shouldn't be necessary for local testing

server/src/external/plex/PlexApiClient.ts
```diff
@@ -1368,12 +1368,24 @@ export class PlexApiClient extends MediaSourceApiClient<PlexTypes> {
     plexEpisode: ApiPlexEpisode,
     mediaLibrary: MediaSourceLibrary,
   ): Result<PlexEpisode> {
-    if (isNil(plexEpisode.duration) || plexEpisode.duration <= 0) {
-      return Result.forError(
-        new Error(
-          `Plex episode ID = ${plexEpisode.ratingKey} has invalid duration.`,
-        ),
-      );
+    let duration = plexEpisode.duration;
```
chrisbenincasa, line 1371: This should be in a separate PR

server/src/external/plex/PlexApiClient.ts
```diff
@@ -1539,12 +1551,24 @@ export class PlexApiClient extends MediaSourceApiClient<PlexTypes> {
     plexMovie: ApiPlexMovie,
     mediaLibrary: MediaSourceLibrary,
   ): Result<PlexMovie> {
-    if (isNil(plexMovie.duration) || plexMovie.duration <= 0) {
-      return Result.forError(
-        new Error(
-          `Plex movie ID = ${plexMovie.ratingKey} has invalid duration.`,
-        ),
-      );
+    let duration = plexMovie.duration;
```
chrisbenincasa, line 1554: This is duplicated to the above - pull it out to a separate method and apply to all media types, not just movies and episodes.

server/src/tasks/fixers/ReconcilePlexDuplicateProgramsFixer.ts
```diff
@@ -0,0 +1,94 @@
+import { inject, injectable } from 'inversify';
```
chrisbenincasa, line 1: Another scope creep - separate PR please.

server/src/tasks/fixers/ReconcilePlexDuplicateProgramsFixer.ts
```diff
@@ -0,0 +1,94 @@
...
+    @inject(KEYS.Database) private db: Kysely<DB>,
+    @inject(KEYS.ProgramDB) private programDB: IProgramDB,
+    @inject(KEYS.ChannelDB) private channelDB: IChannelDB,
+  ) {
+    super();
+  }
+
+  protected async runInternal(): Promise<void> {
+    const duplicateGroups = await sql<DuplicateRow>`
+      SELECT canonical_id AS canonicalId,
```
chrisbenincasa, line 34: why not build this with the ORM? would be easier to detect anything that breaks / changes in future updates?

server/src/db/converters/ProgramMinter.ts
```diff
@@ -99,8 +99,9 @@ export class ProgramDaoMinter {
     movie: MediaSourceMovie,
     localFolderId?: string,
     now: number = +dayjs(),
+    programUuid?: string,
```
chrisbenincasa, line 102: At this point I think an options object would be easier, since multiple default params makes call sites awkward

server/src/services/plex/PlexProgramIdentityService.ts
```diff
@@ -0,0 +1,80 @@
...
+};
+
+export type ResolveExistingProgramResult = {
+  existing: ProgramCanonicalIdLookupResult;
+  reason: 'rating_key' | 'plex_guid' | 'canonical_id';
+  ratingKeyChanged: boolean;
+};
+
+export class PlexProgramIdentityService {
+  constructor(private readonly programDB: IProgramDB) {}
```
chrisbenincasa, line 37: This should use dependency injection like everything else in Tunarr

server/src/services/plex/PlexProgramIdentityService.ts
```diff
@@ -0,0 +1,80 @@
...
+  mediaSource: MediaSourceOrm;
+};
+
+export type ResolveExistingProgramResult = {
+  existing: ProgramCanonicalIdLookupResult;
+  reason: 'rating_key' | 'plex_guid' | 'canonical_id';
+  ratingKeyChanged: boolean;
+};
+
+export class PlexProgramIdentityService {
```
chrisbenincasa, line 36: mark `@injectable`

server/src/services/plex/PlexProgramIdentityService.ts
```diff
@@ -0,0 +1,80 @@
...
+  ): Promise<Maybe<ResolveExistingProgramResult>> {
+    if (req.existingByRatingKey) {
+      return {
+        existing: req.existingByRatingKey,
+        reason: 'rating_key',
+        ratingKeyChanged: false,
+      };
+    }
+
+    const plexGuid = extractPlexGuid(req.incomingMovie.identifiers);
```
chrisbenincasa: Mentioned in the issue, but this should also consider the external provider ids: IMDB, TMDB, and TVDB. The implementation there can be generalized since we'll want to do the same for Jellyfin / Emby / and Local sources as well.

server/src/services/health_checks/PlexIdentityDesyncHealthCheck.ts
```diff
@@ -0,0 +1,36 @@
...
+import { KEYS } from '../../types/inject.ts';
+import {
+  type HealthCheck,
+  type HealthCheckResult,
+  HealthyHealthCheckResult,
+  healthCheckResult,
+} from './HealthCheck.ts';
+
+@injectable()
+export class PlexIdentityDesyncHealthCheck implements HealthCheck {
```
chrisbenincasa, line 13: I think this should be a separate PR

The error message also references Plex specifically but this could be true for items from other sources as well. 

I'm also unsure of the utility of this.... we have the trash page to show this. At the very least the health check should link to the trash page. Not all items that are missing can possibly be reconciled...sometimes people actually delete things.

server/src/services/scanner/MediaSourceMovieLibraryScanner.ts
```diff
@@ -108,7 +111,13 @@ export abstract class MediaSourceMovieLibraryScanner<
     });
 
     return Result.attemptAsync(() =>
-      this.scanSingleInternal(ctx, apiMovie.get(), existingMovie),
+      this.scanSingleInternal(
+        ctx,
+        apiMovie.get(),
+        existingMovie,
+        undefined,
+        new PlexProgramIdentityService(this.programDB),
```
chrisbenincasa, line 119: This should be DI'd

scripts/notarize-macos-dmg.sh
chrisbenincasa, line 1: I'd rather not have duplicate versions of the real scripts that need to be kept in sync with one another. This should be a separate PR anyway. If we _really_ want this, we should amend the existing scripts to be able to run locally. 

Alternatively, it's generally super niche to have to run the fully signed app locally for macOS. If you're testing, you can just run the app directly. If you really need to test something with the native wrapping harness, Xcode can be used. What is the purpose of doing this?

server/src/services/scanner/MediaSourceScanner.ts
```diff
@@ -135,6 +135,9 @@ export abstract class MediaSourceScanner<
       });
 
       await this.mediaSourceDB.setLibraryLastScannedTime(library.uuid, dayjs());
+    } catch (e) {
+      this.logger.error(e, 'Error scanning library %s — lastScannedAt not updated', library.uuid);
```
chrisbenincasa, line 139: Another tangential change that is not related to item resolution - should be a separate PR where we can discuss the implications.

server/src/stream/ProgramStreamDetailsFetcher.ts
```diff
@@ -199,14 +203,20 @@ export class ProgramStreamDetailsFetcher {
       return Result.success({ streamDetails, streamSource });
     } else {
       const filePath = head(firstVersion.mediaFiles)?.path;
-      const serverPath = // details.serverPath ??
-        program.externalIds.find(
-          (eid) => eid.sourceType === server.type,
-        )?.externalFilePath;
+      const plexExternalId = program.externalIds.find(
```
chrisbenincasa, line 206: Another unrelated change; please keep the PR focused.

server/src/stream/ProgramStreamDetailsFetcher.ts
```diff
@@ -216,33 +226,38 @@ export class ProgramStreamDetailsFetcher {
     server: MediaSourceWithRelations,
     potentialFilePath: Nilable<string>,
     serverPath: Nilable<string>,
+    preferredDirectPath?: Nilable<string>,
   ): Promise<StreamSource> {
-    if (isNonEmptyString(potentialFilePath)) {
-      if (await fileExists(potentialFilePath)) {
+    const diskCandidates = [preferredDirectPath, potentialFilePath].filter(
```
chrisbenincasa, line 231: I don't think we should change this in this way without much more discussion. This is a critical path and I think there's some misunderstanding of it here.

server/src/stream/ProgramStreamDetailsFetcher.ts
```diff
@@ -199,14 +203,20 @@ export class ProgramStreamDetailsFetcher {
...
-          (eid) => eid.sourceType === server.type,
-        )?.externalFilePath;
+      const plexExternalId = program.externalIds.find(
+        (eid) => eid.sourceType === server.type,
+      );
+      const serverPath = plexExternalId?.externalFilePath;
+      const directFromDb = plexExternalId?.directFilePath;
+      const preferDirect =
+        server.type === 'plex' &&
+        this.settingsDB.plexSettings().streamPath === 'direct';
```
chrisbenincasa, line 213: Mentioned this should all be split out, but this setting is deprecated and cannot be directly set in the UI anymore. Path replacements are the canonical way of getting direct streaming working. Direct is preferred _always_ over network, even if no replacements are configured, if the file is visible and readable by Tunarr.

server/src/stream/ProgramStreamDetailsFetcher.ts
```diff
@@ -199,14 +203,20 @@ export class ProgramStreamDetailsFetcher {
...
       const filePath = head(firstVersion.mediaFiles)?.path;
-      const serverPath = // details.serverPath ??
-        program.externalIds.find(
-          (eid) => eid.sourceType === server.type,
-        )?.externalFilePath;
+      const plexExternalId = program.externalIds.find(
+        (eid) => eid.sourceType === server.type,
+      );
+      const serverPath = plexExternalId?.externalFilePath;
+      const directFromDb = plexExternalId?.directFilePath;
```
chrisbenincasa, line 210: If this is diverging from the path from the program_version in Tunarr then that is an issue.

chrisbenincasa (changes requested):

server/src/db/channel/LineupRepository.ts
```diff
@@ -487,6 +487,55 @@ export class LineupRepository {
...
+      if (changed) {
+        await this.saveLineup(channelId, { ...lineup, items: newLineupItems });
+      }
+    }
+
+    const channelRows = await this.drizzleDB.query.channelPrograms.findMany({
+      where: (cp, { eq }) => eq(cp.programUuid, oldProgramUuid),
+    });
+
+    for (const row of channelRows) {
```
chrisbenincasa, line 518: This should all be in a transaction

server/src/services/scanner/MediaSourceMovieLibraryScanner.ts
```diff
@@ -206,24 +229,52 @@ export abstract class MediaSourceMovieLibraryScanner<
...
       );
       return;
     }
 
+    const plexLocation = fullMovie.mediaItem?.locations?.find(
+      (loc) => loc.sourceType === mediaSource.type,
+    );
+
+    const reuseProgramUuid =
+      resolved && (await identityService.reconcileRatingKeyIfChanged(
```
chrisbenincasa: I think this is much cleaner and less brittle with if/else

server/src/services/health_checks/PlexIdentityDesyncHealthCheck.test.ts
```diff
@@ -0,0 +1,35 @@
+import { describe, expect, it, vi } from 'vitest';
+import { PlexIdentityDesyncHealthCheck } from './PlexIdentityDesyncHealthCheck.ts';
+
+describe('PlexIdentityDesyncHealthCheck', () => {
+  it('returns healthy when no missing programs in lineups', async () => {
+    const db = {
```
chrisbenincasa, line 6: The tests here mock the DB, which effectively means we're testing the mock and not the actual health check - refactoring the query at all would mean the tests continue to the pass

server/src/db/ProgramDB.ts
```diff
@@ -73,6 +74,7 @@ export class ProgramDB implements IProgramDB {
     private readonly searchRepo: ProgramSearchRepository,
     @inject(KEYS.ProgramStateRepository)
     private readonly stateRepo: ProgramStateRepository,
+    @inject(KEYS.ChannelDB) private readonly channelDB: IChannelDB,
```
chrisbenincasa, line 77: I think this is a smell and indicates that the reconciliation logic should live in its own service or command

chrisbenincasa:
Please lmk if you plan on addressing the review here otherwise I'll patch this and continue on my end

PR #1972, opened 2026-07-29
fix(server): skip scanning virtual placeholder programs in Jellyfin libraries to prevent "no media streams!" warnings

RichardMartija:


server/src/external/jellyfin/JellyfinApiClient.ts
```diff
@@ -485,29 +485,34 @@ export class JellyfinApiClient extends MediaSourceApiClient<JellyfinItemTypes> {
...
           },
-          contributingArtistIds: extraParams.contributingArtistIds?.join(','),
-          excludeItemIds: extraParams.excludeItemIds?.join(','),
-          albumArtistIds: extraParams.albumArtistIds?.join(','),
-        },
-        (v) => isNil(v) || (!isNumber(v) && isEmpty(v)),
-      ),
+          (v) => isNil(v) || (!isNumber(v) && isEmpty(v)),
+        ),
+        excludeLocationTypes: 'Virtual', // prevent pulling in placeholders for missing episodes
```
chrisbenincasa: Is there a reason this is structurally outside omit block? I realize it'll never be omitted but it makes the diff messy

RichardMartija: It makes the intent of the code easier to understand vs putting everything in the omit. You don't read diffs as often as you read code so clean code beats a clean diff IMO. If there are other baseline params in the future it will be nicer to have them separate.

chrisbenincasa: In that case we should move the other constants out in this change

chrisbenincasa: To be fair - the counter here is that if this param ends up to be variable it needs to be moved back into the omit block. Extra cognitive load for the programmer always loses to subjective cleanliness

RichardMartija: I put it with the rest since we may want config for that field it at some point.

PR #1982, opened 2026-08-08
fix(api): contain troubleshoot file reads within session temp directory

bunlongheng:
The troubleshoot API routes build file paths from req.params.file using path.join without checking whether the result stays inside the session temp directory. A request with a traversal sequence like ../../../etc/passwd in the file param would read arbitrary files from the host.

Added path.resolve + startsWith checks in both the /troubleshoot/:id/:file and /troubleshoot/stream/:sessionId/:file handlers that reject any path escaping the intended subdirectory. The validated path is reused for subsequent reads, removing the duplicate join calls.

### DIFF
--- a/server/src/api/streamApi.ts
+++ b/server/src/api/streamApi.ts
@@ -16,7 +16,7 @@ import dayjs from 'dayjs';
 import type { FastifyReply } from 'fastify';
 import { isArray, isNil, isNumber, isUndefined } from 'lodash-es';
 import fs from 'node:fs/promises';
-import { join } from 'node:path';
+import { join, resolve, sep } from 'node:path';
 import { PassThrough } from 'node:stream';
 import { format } from 'node:util';
 import { match } from 'ts-pattern';
@@ -327,13 +327,19 @@ export const streamApi: RouterPluginAsyncCallback = async (fastify) => {
       session.onSegmentRequested(req.ip, req.params.file);
 
       if (req.params.file.endsWith('.vtt')) {
-        const filePath = join(session.workingDirectory, req.params.file);
+        const filePath = resolve(session.workingDirectory, req.params.file);
+        if (!filePath.startsWith(session.workingDirectory + sep)) {
+          return res.status(400).send('Invalid file path');
+        }
         const content = await fs.readFile(filePath, 'utf-8');
         return res.type('text/vtt').send(injectTimestampMap(content));
       }
 
       if (req.params.file.endsWith('.m3u8')) {
-        const filePath = join(session.workingDirectory, req.params.file);
+        const filePath = resolve(session.workingDirectory, req.params.file);
+        if (!filePath.startsWith(session.workingDirectory + sep)) {
+          return res.status(400).send('Invalid file path');
+        }
         const content = await fs.readFile(filePath);
         return res.type('application/vnd.apple.mpegurl').send(content);
       }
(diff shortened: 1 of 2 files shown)


chrisbenincasa:
Have you confirmed the vuln with a reproduction? I think the fix is valid but the threat is slightly more contained than as described. 

Few thoughts:
- I think the path would have to be passed as percent encoded and would otherwise be rejected at route time
- There has to be an active troubleshooting session in order for the paths to hit
- Only the raw read file paths are exposed because sendFile will reject anything above root by default 

Point 2 is important because the raw read paths are already gated by suffix checks. 

Lastly, we may want/need to do the same thing in the main streaming endpoints as a precaution. I'll probably take this PR as is but my top level understanding is that the potential threat here is extremely limited (unless you have a repro to prove otherwise)

chrisbenincasa (changes requested):
Before I approve, I'm just curious if this case actively reproduces. Should also include the same changes to `streamApi` if so.

As an aside for myself, I think I need to invest some time in upgrading the version of `fastify-static` we use to close some other potential holes.

chrisbenincasa:
hey there @bunlongheng - just wondering if you want to follow through here.

bunlongheng:
Hey @chrisbenincasa, really appreciate you taking the time to dig into this so carefully - your breakdown is spot on, and honestly defense-in-depth is the fairer framing here, not a critical read primitive:

- You're right on the routing: an un-encoded `../` gets normalized/rejected before it ever reaches the handler, so a real request would have to percent-encode the traversal.
- Agreed it's gated behind an active troubleshoot session, so it's not something an anonymous caller can reach.
- And great catch that the raw `fs.readFile` branches only fire when `req.params.file` ends in `.m3u8`/`.vtt` - everything else falls through to `sendFile`, which already blocks escapes above root. So the practical worst case is leaking files that happen to end in those suffixes, not an unrestricted `/etc/passwd` read. That's narrower than my original writeup implied, and I think your read is the accurate one.

I don't have a full end-to-end repro built against a live session, so I don't want to overstate the impact. The way I see it, the value is just closing the `join`-without-`resolve` gap so it can't quietly widen later - say a future branch adds a raw read for another suffix. Does that line up with how you're thinking about it?

On the main streaming endpoints - that's a really good call. I'd be happy to extend the same `resolve` + prefix guard into `streamApi.ts` in this PR so it stays consistent, or keep this one scoped and send it as a follow-up. What do you think works best for you? Whichever you prefer, I'm glad to push it.

bunlongheng:
Went ahead and added it to this PR (`0e3753b`) so it's all in one place - the same `resolve` + prefix guard now covers the `.vtt` and `.m3u8` read branches in `streamApi.ts`'s `/stream/channels/:id/:sessionType/:file` handler. Same scope caveats as the troubleshoot routes: it's gated behind an active session and the raw reads are suffix-limited, so this is consistency hardening rather than a critical read - but it closes the same `join`-without-`resolve` gap. `sendFile` was already safe, so I left it as-is.

Happy to adjust the error shape or split it back out into its own PR if you'd rather keep this one narrowly scoped - whatever's easiest for you to review. Thanks again for the thoughtful feedback on this.

chrisbenincasa:
Also note - for the future, it is probably better to first file this as a security vulnerability (https://github.com/chrisbenincasa/tunarr/security/advisories) so we can assess how serious it is + collab on a fix without publicizing the vuln before a fix is in place.

PR #1986, opened 2026-08-15
feat: accelerate VAAPI watermark overlays

NickBenthem:
## Summary

- use `overlay_vaapi` for non-animated still-image channel watermarks when the installed FFmpeg exposes the filter
- keep the main video on VAAPI surfaces and upload only the scaled RGBA watermark
- trim finite-duration watermark inputs and pass through the main video after EOF
- retain the existing software path when hardware filters are disabled, subtitles need a software overlay, the driver is incompatible, the watermark is animated, or `overlay_vaapi` is unavailable

## Validation

- `pnpm lint-changed`
- `pnpm turbo typecheck --filter=@tunarr/server`
- full server suite: 1,141 passed, 2 skipped
- live Intel N100 / FFmpeg 7.1.1 validation on a 1080p VAAPI channel, including a five-second watermark cutoff and a graph with no full-frame `hwdownload`

### DIFF
--- a/server/src/ffmpeg/builder/filter/vaapi/OverlayWatermarkVaapiFilter.ts
+++ b/server/src/ffmpeg/builder/filter/vaapi/OverlayWatermarkVaapiFilter.ts
@@ -1,0 +1,24 @@
+import { OverlayWatermarkFilter } from '@/ffmpeg/builder/filter/watermark/OverlayWatermarkFilter.js';
+import { PixelFormatUnknown } from '@/ffmpeg/builder/format/PixelFormat.js';
+import type { FrameState } from '@/ffmpeg/builder/state/FrameState.js';
+import type { FrameSize } from '@/ffmpeg/builder/types.js';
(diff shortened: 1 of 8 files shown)


NickBenthem:
Apologies, I opened too early. Closing and will open again.

server/src/ffmpeg/builder/filter/vaapi/OverlayWatermarkVaapiFilter.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  nextState(currentState: FrameState): FrameState {
+    return currentState.updateFrameLocation(FrameDataLocation.Hardware);
+  }
+
+  public get filter() {
+    const durationOptions =
+      this.watermark.duration > 0 ? ':eof_action=pass:repeatlast=0' : '';
+    return `overlay_vaapi=${this.getPosition()}${durationOptions}`;
```
chrisbenincasa: we should apply a pixel format filter to change the input to `bgra` before uploading to ensure we preserve transparency

PR #1987, opened 2026-08-15
feat: add current program title overlays

NickBenthem:
Depends on #1986. This is the second PR in the stack; its hardware-overlay commit will disappear from this diff when #1986 merges.

## Summary

- add **Current program title** as a channel overlay source alongside the existing image source
- format episodes as `Show · S01E02 · Episode title`, with sensible movie, video, and track fallbacks
- generate a small transparent lavfi/drawtext input per program, using a cache-backed `textfile` and `expansion=none` so metadata punctuation and percent signs remain data
- reuse the existing position, width, margin, opacity, fade, and duration controls; selecting program titles defaults an untouched duration to 5 seconds and width to 75%
- fail open by skipping the title overlay if its metadata asset cannot be created

## Validation

- `pnpm lint-changed`
- server and web typechecks
- full server suite: 1,141 passed, 2 skipped
- live Intel N100 / FFmpeg 7.1.1 validation on the Simpsons channel: title visible at 2 seconds, absent at 9 seconds, `overlay_vaapi` active, and no full-frame `hwdownload`

### DIFF
--- a/docs/configure/channels/watermarks.md
+++ b/docs/configure/channels/watermarks.md
@@ -6,6 +6,15 @@ Channels can have watermarks to aid in recreating a classic TV experience.
 
 There are many ways to customize watermarks for a channel. Here are some details on specific options:
 
+## Program title overlays
+
+Set the overlay source to **Current program title** to show program metadata
+instead of an image. Episodes are formatted as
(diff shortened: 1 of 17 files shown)


NickBenthem:
Apologies, I opened too early. Closing and will open again.

PR #1993, opened 2026-08-22
fix: resolve external subtitles for Jellyfin and Emby sources

### LINKED ISSUE #1989: [Bug]: External Subtitles Are Not Picked Up When Scanning Emby Libraries
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [Discord](https://discord.gg/scMJxnPXaB) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/chrisbenincasa/tunarr/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I'm using an up to date version of Tunarr; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.

### Description of the bug

When using Emby as a media source external SRT-based subtitles are not picked up by Tunarr when doing a library scan. 

### Reproduction steps

1. Add an Emby Media Source
2. Ensure SRT subtitles exist for the item, and are picked up by Emby
3. Force a library scan
4. View Stream Info for item and observe that SRT subtitles are not listed but any embedded subtitles are

### What is the current _bug_ behavior?

SRT-based subtitles are not picked up in library scan.

### What is the expected _correct_ behavior?

SRT-based subtitles should be available in the stream info for selection.

### Tunarr version

Latest

### FFMPEG encoder type

None

### Specify full Edge version id

_No response_

### Specify Alpha release number

_No response_

### Specify version number

_No response_

### What browser did you experience the the problem on? Please provide if the issue is related to the Web UI and definitely not the server.

_No response_

### Deployment Type

None

### What operating system are you using?

_No response_

### Full server logs

```shell
Too long to post.
```

chrisbenincasa:
## The bug

External (sidecar) subtitles from Jellyfin and Emby were silently dropped before they ever reached ffmpeg. Three problems stacked on top of each other:

- `ProgramStreamDetailsFetcher` never surfaced `external_subtitles` media streams, and the `sidecar` rows it *did* read carried no usable location.
- Jellyfin and Emby report external streams **ahead of** the container's own streams, so `streamIndexOffset` collapsed every external subtitle to index `0`. The index could not address the file even in principle — verified against a real `/Items` payload for a movie with two English SRT sidecars, where both landed on `0`, colliding with each other and with the video stream.
- As a consequence `MediaSourceScanner`'s download pass was dead for these sources: nothing was ever persisted for it to fetch, and the `streamIndex: null` it would have sent addressed the wrong stream anyway.

## The fix

Subtitle location is split into three columns that each say what can actually be done with the value:

| column | meaning |
| --- | --- |
| `path` | a file Tunarr can open right now |
| `sourcePath` | where the media source says the file lives, in *its* namespace — usable only after path replacements, and only when Tunarr sees the same storage |
| `sourceKey` | the source-relative route that serves the file over HTTP (Plex stream key, Jellyfin/Emby delivery URL) |

External subtitles are resolved to a local file **during scanning**, in two steps: shared storage first (no copy, and the only route for image-based external subs), download second. Playback therefore never waits on the media server, and the HTTP-at-playback fallback is gone entirely — `FfmpegStreamFactory`'s `http` subtitle branches were verified unreachable and removed.

`SubtitleExtractorTask`'s hourly guide sweep now tops up sidecars that are still unresolved, so a subtitle missed by a scan — unreachable server, deleted cache file, or a library scanned by an older Tunarr — recovers without a full rescan.

## Also fixed along the way

- **`IsHearingImpaired` / `IsForced` were dropped** during Jellyfin/Emby canonicalization. Both feed stream-selection rules, and `sdh` is frequently the *only* field distinguishing two same-language sidecars.
- **Subtitle cache collisions** between same-codec external subtitles; the source-relative key now discriminates them.
- **Unredacted tokens in logged stream URLs** — `X-Emby-Token=` and `api_key=` query parameters were never scrubbed, which also leaked the token on the Emby *video* URL.
- **`upsertSubtitles` compared `existing.codec` against `incoming.subtitleType`**, so a codec change never invalidated an already-extracted subtitle.

## Behavior change worth calling out

The `enableSubtitleExtraction` gate moved from `SubtitleExtractorTask.runInternal` down into `handleProgram`. External subtitles are files we *fetch*, not streams we transcode out of a container, so they are topped up whether or not extraction is enabled — meaning the hourly guide sweep now runs even with that setting off (it was previously a no-op). Channel-level `subtitlesEnabled` still gates everything.

## Migration

`0045_luxuriant_namora.sql` adds `source_path` / `source_key` and backfills Plex sidecar rows, whose stream key used to live in `path` (which now only ever holds a path Tunarr can open). Rows whose download had already succeeded hold a cache path and are left alone. Registered in `DirectMigrationProvider` as `migration1787428546` and confirmed to run.

## Testing

- `server/src/stream/ExternalSubtitleSelection.test.ts` — 12 tests wiring the real fetcher into the real selector across all three source types.
- `server/src/external/jellyfin/JellyfinApiClient.test.ts` — 7 tests against a real captured `/Items` payload, pinning the offset behavior and both recorded locations.
- `server/src/tasks/SubtitleExtractorTask.test.ts` — 10 tests for the hourly top-up: shared-storage preference, first download, re-download after cache deletion, per-source client dispatch, failed download, and the `subtitlesEnabled` gate.
- `server/src/util/subtitles.test.ts` — 8 tests for delivery-path construction and cache-key discrimination.
- 2 added to `ProgramDB.test.ts` for extracted-subtitle invalidation.

`pnpm turbo typecheck` clean, 1170 tests pass, `pnpm lint-changed` clean.

## Note on the base branch

Opened against `main` rather than `dev` at @chrisbenincasa's request, as an exception to the usual feat/fix split.

https://claude.ai/code/session_0135kZd7grxgAAuhKpqeQqG1


chrisbenincasa:
Reviewed against the head commit (`7795280`): read the full diff, traced the runtime paths, and ran the new suites plus typecheck/lint. The three-column design, the scanner-side resolution, the migration, and the incidental fixes are all solid — but there is a blocking bug in the Jellyfin delivery path.

## Blocking: Jellyfin `sourceKey` is missing the `startPositionTicks` path segment

`externalSubtitleDeliveryPath` (`server/src/util/subtitles.ts`) builds:

```
/Videos/{itemId}/{sourceId}/Subtitles/{index}/Stream.{ext}
```

That is Emby's route shape, not Jellyfin's. The repo's own working `JellyfinApiClient.getSubtitles` (`server/src/external/jellyfin/JellyfinApiClient.ts`) uses:

```
/Videos/{itemId}/{mediaItemId}/Subtitles/{streamIndex}/{tickOffset}/Stream.{ext}
```

Jellyfin requires the `{startPositionTicks}` segment between the index and `Stream.{ext}` (the deprecated no-ticks route is `/Subtitles/{index}/{format}`, also not this shape), so `/Videos/…/Subtitles/0/Stream.srt` matches neither route and will 404.

The new `JellyfinApiClient.test.ts` asserts this URL as the *expected* value, and the captured fixture `movie-with-external-subtitles.json` has no `DeliveryUrl` on the external streams (it appears 0 times in the file), so the fixture itself exercises the broken fallback. The fix is small, but please verify against a live Jellyfin instance rather than the synthetic expectation — either emit the ticks segment (e.g. `/Subtitles/${index}/0/Stream.${ext}`, matching `getSubtitles`) or have the Jellyfin canonicalization pass a server-produced value.

## Secondary (not blocking)

1. `externalSubtitleDeliveryPath` is shared by Jellyfin and Emby, but the two servers' subtitle routes differ — that conflation is the root of the bug above.
2. Emby `DeliveryUrl` is typed `z.string()` and takes the first branch directly; if it is ever a full URL rather than a source-relative path, `getSubtitlesByPath`'s `/` + `trimStart` + `baseURL` join would mangle it. There is no Emby fixture pinning this — worth a check against a real Emby server.
3. The `0045` backfill only matches `path LIKE '/library/%'`; Plex sidecars whose key was an absolute HTTP URL stay in `path` (harmless and pre-existing, but the backfill is slightly incomplete).
4. Top-up passes `dbProgram.externalKey` to the downloader while the scanner passes `program.externalKey`. If those ever diverge, top-up writes a second cache file instead of reusing the scanned one (efficiency only, not correctness).

## Verified

- 37/37 new tests pass (built `@tunarr/types` first — the server resolves it from `dist/`).
- Migration `migration1787428546` runs clean; the `/library/%` vs cache-path distinction in the backfill is correct.
- `server` typecheck clean; ESLint clean on the changed non-test files.
- The codec-invalidation fix, token redaction, and cache-key discrimination changes are correct; the removed `http` subtitle branches in `FfmpegStreamFactory` are genuinely unreachable under the new design.

PR #1997, opened 2026-08-23
fix(db): stop a failed lineup read from erasing the lineup

chrisbenincasa:
## The bug

`SchemaBackedDbAdapter.read()` caught every error from the underlying adapter and returned `null`:

```ts
const data = await this.adapter.read().catch((e) => {
  this.logger.error(e);
  return null;
});
```

lowdb's `TextFile.read()` returns `null` **only** for `ENOENT` and throws for everything else, so that catch collapsed *"this file could not be read"* into *"this file does not exist"*.

`LineupRepository` constructs this adapter with a `defaultValue` for every channel, so the `null` did not hit the early return. It reached `parsed = {}`, failed schema validation, took the merge-with-defaults branch, succeeded, set `needsWriteFlush`, and **wrote the defaults back over the file that had just failed to read** — during the failed read itself, not on a later save.

A transient `EACCES`/`EIO`/`EMFILE` therefore erased a channel's programming. The channel still appeared in the UI with an empty schedule, the guide showed nothing, and streams fell back to flex. The only trace was one `debug`-level line.

## The fix

Read errors propagate. `null` (a genuinely absent file) still initialises from defaults exactly as before.

Three behaviours are pinned by tests so this cannot overshoot into "never write defaults":

- an absent file is still initialised from the defaults, and still written
- a valid file is returned untouched, with no write
- a file missing fields is still repaired by merging the defaults in

## Two related fixes in the same path

**The constructor discarded its injected adapter.**

```ts
constructor(..., private adapter: Adapter<string> = new TextFile(filename)) {
  this.adapter = new TextFile(filename);   // overwrites the parameter
}
```

The `adapter` parameter silently did nothing. The sync variant does not have this. Honouring it is also what makes the class testable at all.

**`saveChannelLineupDirect` wrote non-atomically.** A plain `fs.writeFile` truncates before writing, so an interrupted write leaves a zero-length lineup file — which the adapter path above then treated as empty and overwrote with defaults. It now goes through a new `writeFileAtomic` helper that writes a scratch file and renames it into place.

## Testing

- `SchemaBackedJsonDBAdapter.test.ts` (new, 6 tests) — the failed-read case plus the three regression guards above.
- `writeFileAtomic.test.ts` (new, 4 tests) — kept separate from `fsUtil.test.ts` because that file mocks `node:fs/promises` module-wide and these need the real filesystem.
- The atomicity test was verified to be a genuine discriminator: with a plain `fs.writeFile` it fails on both assertions (the write succeeds *and* the destination is clobbered).
- Full monorepo suite: 1467 passed, 2 skipped. Typecheck and lint clean.

## Notes for review

`SchemaBackedDbAdapter` also backs the settings JSON (`SettingsDBFactory.ts:72`). Propagating read errors there is deliberate: the previous behaviour silently reset a user's settings to defaults on a transient read failure. A genuinely missing settings file is unaffected, since that is the `ENOENT`/`null` path.

### DIFF
--- a/server/src/db/json/SchemaBackedJsonDBAdapter.ts
+++ b/server/src/db/json/SchemaBackedJsonDBAdapter.ts
@@ -24,14 +24,15 @@ export class SchemaBackedDbAdapter<T extends z.ZodTypeAny>
   ) {
     this.schema = schema;
     this.path = filename;
-    this.adapter = new TextFile(filename);
   }
 
   async read(): Promise<z.output<T> | null> {
-    const data = await this.adapter.read().catch((e) => {
-      this.logger.error(e);
-      return null;
-    });
+    // A rejection here means the file exists but could not be read (EACCES,
+    // EIO, EMFILE...). It must propagate: treating it as "no data" sends us
+    // down the merge-with-defaults path below, which flushes those defaults
+    // back to disk and destroys the file we failed to read. Only a genuinely
+    // absent file yields null -- TextFile maps ENOENT, and nothing else, to it.
+    const data = await this.adapter.read();
 
     if (data === null && this.defaultValue === null) {
       this.logger.debug('Unexpected null data at %s', this.path.toString());
--- a/server/src/util/fsUtil.ts
+++ b/server/src/util/fsUtil.ts
@@ -1,4 +1,4 @@
-import fs, { constants } from 'node:fs/promises';
+import fs, { constants, type FileHandle } from 'node:fs/promises';
 import path from 'node:path';
 import { isNodeError } from './index.js';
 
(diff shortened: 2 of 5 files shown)


chrisbenincasa (commented):
## Review: `fix(db): stop a failed lineup read from erasing the lineup` — ✅ Approve

I verified the data-loss mechanism end-to-end, checked every claim against the code and the installed dependencies, and ran the new tests: **10/10 pass** (Node 22), lint and typecheck clean.

### The bug is real and precisely described
In lowdb@7.0.1, `TextFile.read()` returns `null` **only** for `ENOENT` and rethrows everything else (`EACCES`/`EIO`/`EMFILE`...). The old catch collapsed those rejections into `null`, which flowed into `parsed = {}` → schema-fail → merge-with-defaults → `needsWriteFlush` → `write(...)` — **writing defaults back over the file that just failed to read**. Since `LineupRepository.getFileDb` always supplies a `defaultValue`, every channel was on the destructive path; `SettingsDBFactory.ts:72` had the identical hazard for settings.

### The fix checks out
- **Error propagation is behavior-preserving where it must be**: genuine `ENOENT` still initializes from defaults (and still writes), a valid file is returned untouched, and a file missing fields is still repaired by merging defaults in — all three pinned by the new regression guards.
- **Propagation is well-contained**: `LineupRepository` has no swallow-and-defaults catch; the startup migrator catches per channel (logs at error level, **preserves the file**, continues); `getFileDb` only caches the `Low` instance *after* a successful read, so retries are clean; the settings startup path is unaffected (the sync adapter already propagated — the async catch was the inconsistency).
- **Constructor fix is purely a testability fix**: both production constructions (`LineupRepository.ts:191`, `SettingsDBFactory.ts:72`) pass no adapter, so default behavior is unchanged.
- **`saveChannelLineupDirect` was indeed the only non-atomic write path**: lowdb's `TextFile.write` goes through `steno` (temp file + rename with a per-file write queue) and `TextFileSync.write` already does temp+rename — so the new helper closes the one gap.

### Notes on `writeFileAtomic` (see inline comments)
No per-destination serialization (latent race if a concurrent caller is ever added — safe today, the only caller is sequential), and no `fsync` before rename (atomic against process interruption, not full power-loss durability; matches `steno`).

### Test quality is excellent
The `FakeTextFile` mirrors lowdb's contract exactly, the permission-based failure test correctly guards the root case, keeping the file separate from `fsUtil.test.ts` deliberately dodges the module-wide `fs` mock, and the atomicity test genuinely discriminates (a `0o500` directory blocks temp-file *creation* but not overwriting an existing file, so plain `fs.writeFile` would indeed fail both assertions).

No blockers — ship it.

server/src/util/fsUtil.ts
```diff
@@ -16,6 +16,29 @@ export async function fileExists(path: string) {
...
+ * Writes to a scratch file in the destination's directory and renames it into
+ * place. A plain write truncates the destination first, so an interrupted one
+ * leaves a zero-length or partially-written file behind; rename(2) is atomic
+ * within a filesystem, so the destination only ever holds the complete old or
+ * the complete new contents.
+ */
+export async function writeFileAtomic(filePath: string, contents: string) {
+  const tempPath = path.join(
+    path.dirname(filePath),
+    `.${path.basename(filePath)}.${process.pid}.tmp`,
```
chrisbenincasa, line 63: Latent race: the scratch name embeds only `process.pid`, so two *concurrent* `writeFileAtomic` calls for the same destination in one process share a temp path and can clobber each other's scratch file mid-write. This is exactly what `steno` (used by lowdb's `TextFile.write`) serializes with its per-file write queue. Safe today — the only caller, `ChannelLineupMigrator`, is sequential — but worth a comment on the helper or a per-destination queue if a concurrent caller is ever added.

server/src/util/fsUtil.ts
```diff
@@ -16,6 +16,29 @@ export async function fileExists(path: string) {
...
+  );
+
+  try {
+    await fs.writeFile(tempPath, contents);
+    await fs.rename(tempPath, filePath);
```
chrisbenincasa, line 76: Two small hardening notes: (1) there's no `fsync` of the temp file before the rename, so this is atomic against *process interruption* (the scenario being fixed) but not against power loss on some filesystems (e.g. ext4 delayed allocation can still leave the destination empty/partial after a crash). `steno` has the same property, so this matches the codebase — just not full crash durability. (2) `rename` replaces the inode, so the destination's previous mode (e.g. `0600`) is replaced by the umask default (`0644`) — low-stakes for lineup JSON, worth knowing.

server/src/db/json/SchemaBackedJsonDBAdapter.ts
```diff
@@ -24,14 +24,15 @@ export class SchemaBackedDbAdapter<T extends z.ZodTypeAny>
...
-    const data = await this.adapter.read().catch((e) => {
-      this.logger.error(e);
-      return null;
-    });
+    // A rejection here means the file exists but could not be read (EACCES,
+    // EIO, EMFILE...). It must propagate: treating it as "no data" sends us
+    // down the merge-with-defaults path below, which flushes those defaults
+    // back to disk and destroys the file we failed to read. Only a genuinely
+    // absent file yields null -- TextFile maps ENOENT, and nothing else, to it.
+    const data = await this.adapter.read();
```
chrisbenincasa, line 35: Confirmed — this is the exact fix. In lowdb@7.0.1, `TextFile.read()` maps only `ENOENT` to `null` and rethrows everything else, so the old `.catch((e) => { ...; return null; })` made a real read failure indistinguishable from a missing file, which is what sent the merge-with-defaults path over the file. Only genuine `ENOENT` still initializes from defaults.

PR #1998, opened 2026-08-23
fix(backup): stop reporting a backup as successful when it was not written

chrisbenincasa:
## The bug

Four defects in the same path. Together they mean a scheduled backup can report success while producing a truncated or absent archive — and then delete the known-good archives that preceded it. Discovered only at restore time.

**1. Destination errors were unobserved.**

```ts
const outStream = createWriteStream(backupFileName);   // no 'error' listener, ever
archive.pipe(outStream);
```

`archive.pipe(out)` does not forward the destination's errors back to the archiver. An `ENOSPC`/`EACCES`/`EROFS` emits `'error'` on a stream with no listener, which is an uncaught exception. `RunServerCommand` catches `uncaughtException`, logs, and does **not** exit — so the process carries on and the promise gated on the archive never settles.

**2. It resolved before the bytes landed.**

```ts
archive.on('end', () => resolve(void 0));
```

The archiver's `'end'` fires when its readable side drains, not when the destination has flushed and closed.

**3. Retention ran before any of that was confirmed.** The ordering was `finalize()` → `rm(tempDir)` → `deleteOldBackupIfNecessary()` → *then* await the archive promise. Old backups were pruned before the new one was known good.

**4. `BackupTask` ignored the failure.** It branched only on `result.type === 'success'`; the `'error'` case fell through with no log, and the task reported success regardless.

## The fix

Stream wiring moves into a new `pipeArchiveToFile` helper that listens on both sides and settles on the **destination's** `'close'`. `backup()` awaits it before doing anything else, prunes only after the archive is confirmed on disk, and returns `{type: 'error'}` on failure. A failed output now fails the task.

## Also: archiver warnings were silently dropping entries

Found while writing the tests. archiver reports a file or directory it could not add as a **`'warning'`**, not an `'error'` — the entry is dropped and the archive completes normally. `backup()` archives directories that may legitimately be absent (`images/`, `cache/`, search snapshots), so this silently produced short backups that still reported success.

`pipeArchiveToFile` now surfaces warnings to a callback, and `backup()` logs them against the backup filename. They are deliberately **not** fatal — a genuinely absent optional directory should not fail the backup, it should just be visible.

## Testing

- `pipeArchiveToFile.test.ts` (new, 4 tests) — rejects on an unwritable destination, rejects on archive error, resolves only once the whole archive is on disk (8MB payload, so resolving on `'end'` would report a truncated file), and reports entries it could not add.
- `BackupTask.test.ts` (new, 2 tests) — asserts through the public `run()` `Result` that a failed backup does not produce a success.
- Full monorepo suite: 1463 passed, 2 skipped. Typecheck and lint clean.

## Note for review

A single failing output now fails the whole task, where previously every failure was swallowed and the task always succeeded. That is the intended change, but it does mean a misconfigured output that was silently failing will start surfacing as a task failure.


chrisbenincasa (commented):
## Review: `fix(backup): stop reporting a backup as successful when it was not written` — ⚠️ Request changes

The four-defect diagnosis is excellent and the new pieces are individually sound, but I found a blocking await-ordering bug in `backup()` that defeats the PR's central promise: a failing destination can hang the backup forever instead of failing it.

### Verified
- **All four defects are real.** (1) `archive.pipe(out)` doesn't forward destination errors and the old write stream had no `'error'` listener → uncaught exception on `ENOSPC`/`EACCES`/`EROFS`; `RunServerCommand` catches `uncaughtException` and `unhandledRejection` (lines 71–77) and logs-and-continues, so the process survives and the promise never settles. (2) archiver's `'end'` fires on the readable-side drain, not the destination flush (confirmed in archiver@7.0.1: `finalize()` resolves on the module's `'end'`). (3) Old ordering ran retention before the archive was confirmed. (4) `BackupTask` branched only on `'success'`.
- **`pipeArchiveToFile` is well-designed in isolation** (dest `'error'` → reject, `'close'` → resolve, archive `'error'` → reject, `'warning'` → callback), and the 4 helper tests are good — especially the 8MB "not truncated" test and the ENOENT-warning test.
- **The `BackupTask` fix is correct**: `Task2.run()` (Task.ts) converts the throw into a failed `Result`, so the public `run()` semantics hold; both tests pass.
- 6/6 new tests pass under Node 22; lint and typecheck clean.

### ⛔ Blocking: `backup()` awaits `archive.finalize()` before `written` (see inline comment)
Reproduced empirically (Node 22, archiver 7.0.1) with `backup()`'s exact ordering: with a realistic multi-entry payload, a destination failure left `await archive.finalize()` **never settling** (the module's `'end'` never fires once the pipe is broken), `written`'s rejection went **unhandled**, and `backup()` never returned — no `{type:'error'}`, no cleanup, no task failure. With a tiny single-entry payload the error path happened to work, so the outcome is a race; real backups (db.db + settings + lineups + images + cache) hit the hang. The new tests miss this because `pipeArchiveToFile.test.ts` tests the helper in isolation and `BackupTask.test.ts` mocks `backup()` entirely — neither exercises the integration ordering.

### Minor issues
- **Partial archive file is left behind on failure** (see inline comment): the archive streams directly to the final destination name and the catch only removes `tempDir`; a truncated `tunarr-backup-*.tar` lingers and the retention glob counts it toward `maxBackups`, so a failed run can trigger premature pruning of good backups on the next success.
- **One failure fails all outputs** in `BackupTask` — acknowledged and defensible, just coarse-grained with multiple outputs.

No blockers beyond the await-ordering issue; please fix the ordering (await `written` first, observe `finalize` without blocking on it) and add an integration test with a failing destination before merging.

server/src/db/backup/ArchiveDatabaseBackup.ts
```diff
@@ -146,22 +147,26 @@ export class ArchiveDatabaseBackup extends DatabaseBackup<string> {
         SearchSnapshotsFolderName,
       )
       .glob('*.xml', { cwd: getDatabasePath('') });
-    await archive.finalize();
+    try {
+      await archive.finalize();
```
chrisbenincasa, line 151: Blocking: on a destination write failure, this `await` can **never settle**. `finalize()`'s promise resolves on the archiver module's `'end'`, which only fires when the archive's readable side is fully consumed. When the write stream errors, Node's `pipe()` unpipes the archiver, and with data still buffered (a real backup has many entries) the module never ends — so `await written` below is never reached, `backup()` hangs forever, and `written`'s rejection is **unhandled** (fired the process `unhandledRejection` handler in my reproduction).

I reproduced this with backup()'s exact ordering (Node 22, archiver 7.0.1): a 4MB two-entry payload against a failing destination hung indefinitely; a tiny 1MB payload happened to take the working path (and in one run archiver's internal queue also threw an uncaught `tar-stream` "already finalized or destroyed" exception). The outcome is a race, and the realistic case is the hang — the exact scenario this PR claims to fix.

Fix: await `written` first (it settles on success and both failure modes) and observe `finalize` without blocking on it:

```ts
const finalize = archive.finalize().catch(() => void 0); // observed, non-blocking
try {
  await written;   // source of truth: dest 'close' after a full drain implies the archive ended
  await finalize;  // already settled on the success path
} catch (e) { ... }
```

(`Promise.all`/`allSettled` won't work — they also gate on `finalize`, which never settles on destination failure.) Please add an integration test that drives backup() (or this exact ordering) with a failing destination and asserts `{type:'error'}`.

server/src/db/backup/ArchiveDatabaseBackup.ts
```diff
@@ -146,22 +147,26 @@ export class ArchiveDatabaseBackup extends DatabaseBackup<string> {
...
+      await written;
+    } catch (e) {
+      this.logger.error(e, 'Error creating backup at %s', backupFileName);
+      await fs.rm(tempDir, { recursive: true, force: true });
```
chrisbenincasa, line 155: Minor: the catch removes `tempDir` but not the partial archive at `backupFileName` (the archive streams directly to the final destination name). A truncated `tunarr-backup-*.tar` lingers in the output directory, and `deleteOldBackupIfNecessary`'s glob (`/tunarr-backup-(\d{8}_\d{6})/`) counts it toward `maxBackups` — so a failed run can trigger premature pruning of good backups on the next successful run. Consider `await fs.rm(backupFileName, { force: true })` here.

server/src/db/backup/ArchiveDatabaseBackup.ts
chrisbenincasa: Fixed in 711f6231, using the ordering you suggested.

`finalize()` is now started and observed without blocking, and `written` is the thing awaited:

```ts
const finalized = archive.finalize().catch(() => void 0);
try {
  await written;
  await finalized;
} catch (e) { ... }
```

There was a second unhandled-rejection window you didn't hit in your repro: the destination can fail while the SQLite backup and search snapshot are still running, well before the `try` block. So `written.catch(() => void 0)` is registered immediately after `pipeArchiveToFile`, which marks it handled without consuming it — `await written` still observes the rejection.

Integration tests added in `server/src/db/backup/ArchiveDatabaseBackup.test.ts`, driving the real `backup()` and asserting `{type: 'error'}`:

- **destination cannot be opened** — a regular file where the output directory should be, so the archive open fails with ENOTDIR.
- **writes fail mid-stream** — the archive path is a symlink to `/dev/full`, so it opens fine and then every write fails with ENOSPC. `test.skipIf(!existsSync('/dev/full'))` so it doesn't break non-Linux.

Both use a multi-megabyte, multi-entry payload for the reason you gave — a tiny payload can finish before the failure lands and take the working path.

I checked they actually catch it: reverting just `ArchiveDatabaseBackup.ts` to the previous ordering and re-running makes **both tests time out at 15s**, i.e. the hang, not the error. With the fix both pass in ~100ms.

server/src/db/backup/ArchiveDatabaseBackup.ts
chrisbenincasa: Good catch — fixed in 711f6231. The partial archive is now removed alongside the temp directory:

```ts
const cleanups = await Promise.allSettled([
  fs.rm(backupFileName, { force: true }),
  fs.rm(tempDir, { recursive: true, force: true }),
]);
```

`allSettled` so a failure removing one doesn't skip the other; rejections are logged rather than thrown, since the backup has already failed and the return value shouldn't change.

One wrinkle worth flagging. On the unopenable-destination path, `fs.rm(backupFileName, { force: true })` itself rejects with ENOTDIR — `force` swallows ENOENT but not ENOTDIR — which logged `Unable to clean up failed backup` for a file that could never have existed. I suppressed that one code via the existing `isNodeError` helper so the warning only fires when there's genuinely something stuck on disk. Slightly beyond what you asked for; happy to drop it if you'd rather keep the hunk minimal.

The mid-stream-failure test asserts the directory is left empty afterward, so the premature-pruning scenario you described is now covered.

PR #2000, opened 2026-08-23
fix(guide): make the guide build retry actually retry, and stop hiding failures

chrisbenincasa:
## The bug

`buildChannelGuideWithRetries` wraps its work in `retry({retries: 15, factor: 2, maxRetryTime: 30000})` — but the callback catches its own errors:

```ts
await retry(async () => {
  try { ... build ... }
  catch (err) { this.logger.error(err, 'Unable to update internal guide data'); }
  finally {
    this.lastUpdateTime[channelId] = this.currentUpdateTime[channelId]!;
    this.lastEndTime[channelId]   = this.currentEndTime[channelId]!;
    this.currentUpdateTime[channelId] = -1;
  }
}, { retries: 15, ... });
```

A callback that never rejects looks like success on attempt 1, so **the 15 retries never happened**.

The `finally` is the more damaging half. It advanced `lastUpdateTime` and `lastEndTime` to the *intended* window end regardless of whether `cachedGuide[channelId]` was ever assigned.

`getChannelLineup` consults `lastEndTime` to decide whether the cached guide covers a requested range. So a failed build told every downstream consumer the guide was fresh when there was no guide at all:

- the "end time exceeds what the current cached guide generation" warning was skipped
- the lookup fell through to `isNil(channelAndLineup) → return`, yielding `undefined`
- the channel rendered as an **empty row** in the web guide and was **omitted from XMLTV**

One `error` line was the only trace, and it said nothing about which consumers were now serving nothing.

## The fix

The `catch` moves outside `retry()` so failures are actually retried, and the timestamps are advanced only when the build produced a guide. Leaving them behind is what makes the failure visible downstream and gets the channel rebuilt next cycle.

The catch is **deliberately not rethrown**. `buildAllChannels` fans out over `Promise.all`, so letting one channel's failure escape would abort every other channel's build and skip the XMLTV write entirely. There's a test for that.

## On the retry ceiling

Retry options move to a `protected readonly` field. Worth being explicit about the trade-off, since this code now retries where it previously did not:

`buildChannelGuide` takes the channel's lock **before** calling this, so the retry sequence is held under that lock. The bound that matters is therefore `maxRetryTime` (30s), not the retry count. I confirmed `maxRetryTime` is genuinely honoured — `retry_operation.js:52` compares elapsed time against it and stops — so 15 retries at `factor: 2` cannot run away; the sequence is capped at 30s.

The lock is per-channel (`MutexMap`), so a failing channel delays only its own rebuilds. Other channels are unaffected.

## Testing

Three tests in the existing `TvGuideService.test.ts`:

- **retries a channel whose guide build fails** — verified red against the original implementation: exactly 1 attempt, `expected 1 to be greater than 1`
- **does not record a channel as built when its build failed** — asserts through the public `getStatus()` that the failed channel is absent from `lastUpdate`
- **still builds and writes the other channels** — the guard on not rethrowing

The test subclass lowers the retry timings; sitting through the 30s production ceiling in a unit test would be silly.

Full monorepo suite: 1460 passed, 2 skipped. Typecheck and lint clean.

## Deliberately not in this PR

The sweep also found that `LineupRepository.loadLineup` returns `Low.data` **by reference**, so a guide build reads the same object a concurrent `saveLineup` mutates — the same shape as febfd898, which its lock does not cover because the conflicting writer is `saveLineup`.

I left it out. The fix has to change aliasing semantics across ~10 call sites, at least one of which (`removeProgramsFromLineup`) mutates the returned object in place and relies on that. Getting it wrong silently drops lineup edits, which does not belong bundled into this PR. It gets its own branch.

### DIFF
--- a/server/src/services/TvGuideService.ts
+++ b/server/src/services/TvGuideService.ts
@@ -133,6 +133,15 @@ const PlaceholderChannelId = v4();
 export class TVGuideService {
   private timer: Timer;
   private locks = new MutexMap();
+
+  // Bounded by maxRetryTime rather than by the retry count: buildChannelGuide
+  // holds this channel's lock for the whole of it, so the ceiling on how long a
+  // failing channel can block its own rebuilds is what matters.
+  protected readonly guideBuildRetryOptions: retry.Options = {
+    retries: 15,
+    factor: 2,
+    maxRetryTime: 30_000,
+  };
   private xmltv: XmlTvWriter;
   private eventService: EventService;
   private programConverter: ProgramConverter;
@@ -150,7 +159,7 @@ export class TVGuideService {
   private accumulateTable: Record<string, number[]> = {};
   private channelsById?: Record<string, ChannelWithLineup>;
 
-  @InjectLogger() private declare readonly logger: Logger;
+  @InjectLogger() declare private readonly logger: Logger;
 
   constructor(
     @inject(XmlTvWriter) xmltv: XmlTvWriter,
@@ -991,36 +1000,43 @@ export class TVGuideService {
       this.currentEndTime[channelId]! - this.currentUpdateTime[channelId]!;
     const dur = dayjs.duration(thisGuideLength).humanize();
     this.logger.debug('Building Channel %s for %s', channelId, dur);
-    await retry(
-      async () => {
-        try {
-          await this.timer.timeAsync(
-            `Built Channel ${channelId} TV Guide for ${dur}`,
-            async () => {
-              this.cachedGuide[channelId] =
-                await this.buildGuideInternal(channelId);
-              if (
-                writeXmlTv &&
-                !this.channelsById![channelId]!.channel.stealth
-              ) {
-                await this.writeXmlTv();
-              }
-            },
-          );
-        } catch (err) {
-          this.logger.error(err, 'Unable to update internal guide data');
-        } finally {
-          this.lastUpdateTime[channelId] = this.currentUpdateTime[channelId]!;
-          this.lastEndTime[channelId] = this.currentEndTime[channelId]!;
-          this.currentUpdateTime[channelId] = -1;
-        }
-      },
-      {
-        retries: 15,
-        factor: 2,
-        maxRetryTime: 30000,
-      },
-    );
+
+    try {
+      // The catch has to live outside retry(): a callback that swallows its own
+      // errors never rejects, so retry() sees success on the first attempt and
+      // the retries never happen.
+      await retry(async () => {
+        await this.timer.timeAsync(
+          `Built Channel ${channelId} TV Guide for ${dur}`,
+          async () => {
+            this.cachedGuide[channelId] =
+              await this.buildGuideInternal(channelId);
+            if (writeXmlTv && !this.channelsById![channelId]!.channel.stealth) {
+              await this.writeXmlTv();
+            }
+          },
+        );
+      }, this.guideBuildRetryOptions);
+
+      this.lastUpdateTime[channelId] = this.currentUpdateTime[channelId]!;
+      this.lastEndTime[channelId] = this.currentEndTime[channelId]!;
+    } catch (err) {
+      // Deliberately leave lastUpdateTime and lastEndTime where they were.
+      // getChannelLineup consults lastEndTime to decide whether the cached
+      // guide covers a requested range, so advancing them for a build that
+      // produced nothing tells every downstream consumer the guide is fresh
+      // when there is no guide at all.
+      this.logger.error(
+        err,
+        'Unable to update guide data for channel %s; giving up after retries',
+        channelId,
+      );
+    } finally {
+      // Not swallowed silently but not rethrown either: buildAllChannels fans
+      // out over Promise.all, so letting this escape would abort every other
+      // channel's build and skip the XMLTV write entirely.
+      this.currentUpdateTime[channelId] = -1;
+    }
   }
 
   private async writeXmlTv() {
--- a/server/src/services/TvGuideService.test.ts
+++ b/server/src/services/TvGuideService.test.ts
@@ -56,8 +56,6 @@ function makeChannelWithLineup(overrides?: Partial<ChannelOrm>) {
   return { channel, lineup: makeEmptyLineup() };
 }
 
-
-
 describe('TVGuideService', () => {
   describe('buildAllChannels', () => {
     it('removes a deleted channel from XMLTV output on the next guide build', async () => {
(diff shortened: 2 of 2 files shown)


server/src/services/TvGuideService.ts
```diff
@@ -991,36 +1000,43 @@ export class TVGuideService {
...
+            this.cachedGuide[channelId] =
+              await this.buildGuideInternal(channelId);
+            if (writeXmlTv && !this.channelsById![channelId]!.channel.stealth) {
+              await this.writeXmlTv();
+            }
+          },
+        );
+      }, this.guideBuildRetryOptions);
+
+      this.lastUpdateTime[channelId] = this.currentUpdateTime[channelId]!;
```
chrisbenincasa, line 1021: Minor new edge worth a follow-up guard: because failures now deliberately leave `lastEndTime` stale, a channel that failed every build since its last success has `lastEndTime` in the past. A subsequent channel-settings change then hits `updateCachedChannel(…, recalculateGuide=true)`, which computes `dayjs.duration(dayjs(end).diff(now))` — **negative** — and calls `refreshGuide(negative duration, force=true)`. In `getChannelPrograms`, `while (currentProgram.startTimeMs < currentEndTimeMs)` never runs, so an **empty guide gets cached and marked built** (until the next full rebuild). This path wasn't reachable before, since failures still advanced the timestamps. Suggest clamping in `updateCachedChannel`: use the fallback duration when `!end || end <= now`.

server/src/services/TvGuideService.ts
```diff
@@ -991,36 +1000,43 @@ export class TVGuideService {
...
-        factor: 2,
-        maxRetryTime: 30000,
-      },
-    );
+
+    try {
+      // The catch has to live outside retry(): a callback that swallows its own
+      // errors never rejects, so retry() sees success on the first attempt and
+      // the retries never happen.
+      await retry(async () => {
```
chrisbenincasa, line 1008: Confirmed — this is the core fix. The red/green check against the original implementation shows exactly 1 attempt before (the swallowed-error callback made retry() see success on attempt 1) vs. genuine retries after. One note: deterministic build failures (e.g. the "General algorithm error" from an inconsistent lineup — the test's own scenario) will burn the full ~30s retry sequence every cycle per broken channel. The per-channel lock contains it, but a `bail()` on non-transient error classes would avoid the burn — a good follow-up, not a blocker.

server/src/services/TvGuideService.ts
```diff
@@ -133,6 +133,15 @@ const PlaceholderChannelId = v4();
 export class TVGuideService {
   private timer: Timer;
   private locks = new MutexMap();
+
+  // Bounded by maxRetryTime rather than by the retry count: buildChannelGuide
+  // holds this channel's lock for the whole of it, so the ceiling on how long a
+  // failing channel can block its own rebuilds is what matters.
+  protected readonly guideBuildRetryOptions: retry.Options = {
```
chrisbenincasa, line 140: Verified the ceiling claim at the library level: `retry` is `async-retry`, which forwards these options to `retry@0.13.1`'s `operation()`. In the installed `retry_operation.js`, `maxRetryTime` is read (line 10) and enforced (line 52, `currentTime - this._operationStart >= this._maxRetryTime`), with `_operationStart` set at the first `attempt()` — so the sequence is bounded at ~30s + one final attempt (≈5–6 attempts with default `minTimeout` 1000ms and async-retry's `randomize: true`), not the ~9 hours of 15 unbounded factor-2 retries. It's also in `@types/retry`'s `WrapOptions`, so typecheck stays clean.

chrisbenincasa (commented):
## Review: `fix(guide): make the guide build retry actually retry, and stop hiding failures` — ✅ Approve

I verified both bugs empirically (red/green against the original implementation), traced the retry library chain to confirm the `maxRetryTime` bound, and checked the lock and downstream semantics. 16/16 tests pass under Node 22; lint and typecheck clean.

### Both bugs are real — and the tests prove it
Running the PR's test file against `main`'s implementation fails exactly the two bug-pinning tests, as claimed:
- *"retries a channel whose guide build fails"* → `expected 1 to be greater than 1` — the callback caught its own errors, so `retry()` saw success on attempt 1 and the 15 retries never happened.
- *"does not record a channel as built when its build failed"* → the old `finally` advanced `lastUpdateTime`/`lastEndTime` even for a failed build, which is what told `getChannelLineup` (it consults `lastEndTime` before falling through to `cachedGuide`) that a nonexistent guide was fresh → empty rows / omitted from XMLTV.
- The third test ("still builds and writes the other channels") passes on `main` too — it's the over-correction guard, as designed.

### The fix is structurally right
The catch moved outside `retry()` so failures actually propagate to the retry loop; timestamps advance only after success; the `finally` only resets `currentUpdateTime`; and the deliberate non-rethrow is justified — `buildAllChannels` fans out over `Promise.all`, so a rethrow would abort the other channels' builds and skip the XMLTV write.

### The retry ceiling claim checks out at the source level
`retry` here is `async-retry`, which forwards options to `retry@0.13.1`'s `operation()`. I confirmed in the installed `retry_operation.js`: `maxRetryTime` is read (line 10) and enforced (line 52, `currentTime - this._operationStart >= this._maxRetryTime`), with `_operationStart` set at the first `attempt()` — so the sequence is genuinely bounded at ~30s + one final attempt (≈5–6 attempts with the default `minTimeout` 1000ms and `randomize: true`), not the ~9 hours that 15 unbounded factor-2 retries would take. It's also typed (`@types/retry` `WrapOptions.maxRetryTime`), so typecheck stays clean.

### Lock and isolation semantics are right
`buildChannelGuide` holds the per-channel `MutexMap` lock across the whole retry sequence; readers (`getChannelLineup`) don't take it; other channels build concurrently under `Promise.all`. A failing channel burns at most ~30s of its *own* rebuild budget.

### Issues / notes (details inline)
1. **(Minor, new edge)** After a failure leaves `lastEndTime` stale (in the past), a later channel-settings change (`updateCachedChannel(…, recalculateGuide=true)`) computes a **negative** duration (`dayjs.duration(dayjs(end).diff(now))`) and can cache an **empty guide as "built"** (`getChannelPrograms`' `while (startTimeMs < currentEndTimeMs)` never runs). Newly reachable because failures no longer advance the timestamps; a `!end || end <= now` fallback guard would close it. See inline comment.
2. **(Nit)** `getChannelLineup`'s "end time exceeds" warning now fires on every consumer request in the stale window of a failed channel (plus one error per cycle after the retry burn) — intended fail-loud behavior, but worth rate-limiting if a persistently broken channel makes it noisy.
3. **(Nit)** Unrelated drive-by: `@InjectLogger() private declare readonly` → `@InjectLogger() declare private readonly` — same cosmetic reorder as #1999; both orders exist in the codebase.
4. **(Note)** Retrying deterministic failures (e.g. the "General algorithm error" from an inconsistent lineup) is pure cost — ~30s per cycle per broken channel, contained by the per-channel lock. A `bail()` on non-transient errors would be a good follow-up.

No blockers — ship it.

PR #2008, opened 2026-08-23
fix: populate scan_kind for interlaced Emby, Jellyfin, and Plex content

coreyjv:
Media source clients never mapped the interlaced/progressive info returned by Emby, Jellyfin, and Plex into MediaItem.scanKind, so it was always stored as 'unknown'.

Users will have to rescan libraries to pick up the change. Results for a library before the fix:

```
sqlite> SELECT scan_kind, COUNT(*) FROM program_version GROUP BY scan_kind;
scan_kind    COUNT(*)
-----------  --------
progressive  8       
unknown      4049    
```

Results after scanning a specific show after the fix:

```
sqlite> SELECT scan_kind, COUNT(*) FROM program_version GROUP BY scan_kind;
scan_kind    COUNT(*)
-----------  --------
interlaced   142     
progressive  13      
unknown      3902
```

### DIFF
--- a/server/src/external/emby/EmbyApiClient.ts
+++ b/server/src/external/emby/EmbyApiClient.ts
@@ -46,6 +46,7 @@ import {
   isNil,
   isNull,
   isNumber,
+  isUndefined,
   mapValues,
   omitBy,
   orderBy,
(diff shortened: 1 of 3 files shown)


server/src/external/emby/EmbyApiClient.ts
```diff
@@ -876,13 +877,10 @@ export class EmbyApiClient extends MediaSourceApiClient<EmbyItemTypes> {
...
 
     while (page * pageSize < totalCount) {
-      const result = await this.getRawItems(
-        null,
-        libraryId,
-        ['BoxSet'],
-        [],
-        { offset: page * pageSize, limit: pageSize },
-      );
+      const result = await this.getRawItems(null, libraryId, ['BoxSet'], [], {
```
chrisbenincasa, line 880: did you run the formatter? or maybe the formatter was never run on this file in the first place... i think the pervious formatting is what is configured in the repo

chrisbenincasa (approved):

PR #2012, opened 2026-08-23
fix(scheduling): stop an out-of-period slot offset from killing the schedule

chrisbenincasa:
## Problem

A time slot's `startTime` is an offset into the schedule period, but nothing enforces it.
One user reported a daily schedule containing a slot at `107400000` — 29h 50m.

`TimeSlotService` reduces its **cursor** modulo the period but never the **slot**:

```ts
const currOffset = totalOffsetMs % periodMs;   // :183
...
if (slot.startTime <= currOffset && currOffset < endTime)   // :210
```

`currOffset` is bounded to `[0, periodMs)` while `slot.startTime` is not, so such a slot
matches no cursor position at all. That asymmetry is the bug.

Measured with the reporter's settings — 5-minute pad, 10-minute lateness:

| Schedule generated at | Result |
|---|---|
| 00:00 – 05:45 | throws `Could not find a suitable slot` — **70 of 288 five-minute marks, 24% of the day** |
| 05:50 – 23:55 | 1 lineup item: a single 48–72 hour flex block, so the channel plays nothing |

A control slot inside the period never throws at any of the 288 start times and produces
32 items with a longest block of 1.5h.

## Fix

Normalize the offset into the period when building the slot list, so schedules already
stored with such a value start working again instead of needing a manual edit.

Normalizing can land two slots on the same offset, which turned out to matter. Equal
`startTime`s give the earlier slot a zero-width window — its `endTime` is the next slot's
identical `startTime` — so it airs nothing, and which of the two died depended on array
order:

| Slots, in array order | Without collision handling |
|---|---|
| A@05:50, B@29:50 | A silently dropped |
| B@29:50, A@05:50 | B silently dropped |

So collisions collapse explicitly, keeping the first — the rule the editor already applies
in `TimeSlotEditorPage.tsx:241-252` when converting a weekly schedule to a daily one. Had
the server kept the last instead, what airs would not match what the editor shows.

## Test plan

- [x] A slot at `t` and one at `t + period` schedule **identically** — previously 33 items or 1
- [x] No throw when generated before the slot's offset
- [x] Real programming rather than one long flex block
- [x] A collided pair produces the same lineup as the surviving slot alone
- [x] Existing scheduling suite unaffected — 124 tests green

The companion input validation is deliberately a separate PR, since it rejects requests the
API currently accepts and may want its own discussion.

### DIFF
--- a/server/src/services/scheduling/TimeSlotService.outOfRange.test.ts
+++ b/server/src/services/scheduling/TimeSlotService.outOfRange.test.ts
@@ -1,0 +1,132 @@
+import { randomUUID } from 'node:crypto';
+import type { TimeSlotSchedule } from '@tunarr/types/api';
+import dayjs from '@/util/dayjs.js';
+import { describe, expect, test } from 'vitest';
+import { createFakeProgramOrm } from '../../testing/fakes/entityCreators.ts';
+import type { SlotSchedulerProgram } from './slotSchedulerUtil.js';
+import { scheduleTimeSlots } from './TimeSlotService.ts';
+
+const ONE_DAY = 86_400_000;
+const TIME_OF_DAY = 21_000_000; // 05:50
+const OUT_OF_RANGE = ONE_DAY + TIME_OF_DAY; // 107_400_000, as reported
+
+const programs: SlotSchedulerProgram[] = Array.from({ length: 8 }, (_, i) => ({
+  ...createFakeProgramOrm({
+    uuid: `mv${i}`,
+    title: `Movie ${i}`,
+    type: 'movie',
+    duration: 90 * 60 * 1000,
+  }),
+  parentFillerLists: [],
+  parentCustomShows: [],
+  parentSmartCollections: [],
+}));
+
+// The settings from the bug report.
+function makeSchedule(startTime: number): TimeSlotSchedule {
+  return {
+    type: 'time',
+    flexPreference: 'distribute',
+    maxDays: 1,
+    padMs: 5 * 60 * 1000,
+    latenessMs: 10 * 60 * 1000,
+    period: 'day',
+    timeZoneOffset: 0,
+    slots: [
+      {
+        id: randomUUID(),
+        startTime,
+        type: 'movie' as const,
+        order: 'next' as const,
+        direction: 'asc' as const,
+      },
+    ],
+  };
+}
+
+describe('time slots with a startTime beyond the period', () => {
+  const midnight = dayjs().startOf('day');
+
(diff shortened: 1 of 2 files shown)


server/src/services/scheduling/TimeSlotService.outOfRange.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      ),
+    ]);
+
+    expect(outOfRange.lineup.length).toBe(inRange.lineup.length);
+    expect(outOfRange.lineup.map((i) => i.duration)).toEqual(
+      inRange.lineup.map((i) => i.duration),
+    );
+  });
+
+  test('collapses two slots that normalize onto the same offset, keeping the first', async () => {
```
chrisbenincasa, line 103: Optional test strengthening: this documents the collision rule but doesn't actually guard it. Both slots are identical `movie`/`next` slots over the same program pool, whose iterators are deterministic (`ContentProgramOrderedIterator`), so the lineup is identical no matter which of the two survives — this test even passes against pre-fix code (it was the 1 of 4 that stayed green). A regression to keep-last, or removal of the `uniqBy` dedupe, would not be caught. To pin the documented keep-first rule, make the second slot a different type (e.g. a `show` slot over its own program pool) and assert the collided lineup equals the lineup of the single *first* slot.

chrisbenincasa (commented):
Review notes — LGTM overall. Verified locally against the PR head.

**Verification**
- Ran the new suite against pre-fix code: 3/4 fail (throw case, long-flex-block case, equivalence case) — the regression tests genuinely capture the bug; 4/4 pass on this branch.
- Full scheduling suite: 124/124 green.
- Traced the old `nextPeriodOffset` branch for the reported settings — confirms the ~46h flex block (22:50 + 24:00) and the pre-05:50 throw described in the PR body.
- Confirmed the keep-first dedupe matches the editor's weekly→daily conversion (`groupBy` + `mapValues(head)` in `TimeSlotEditorPage.tsx`).
- Confirmed slot/filler iterator construction doesn't read `startTime`, so normalizing only `sortedSlots` is safe.

**Notes (non-blocking)**
- The collision test below doesn't actually pin keep-first (see inline comment).
- Collision dedupe silently drops the later slot of any type pair (e.g., flex@29:50 + movie@05:50) based purely on array order — consistent with the editor, but the server never surfaces a "slot dropped" signal.
- Deferring input validation to a separate PR is the right call; runtime normalization also repairs already-persisted broken schedules, which save-time validation alone wouldn't.

PR #2035, opened 2026-08-26
fix(server): restore FFmpeg reset response schema

### LINKED ISSUE #2029: Typo 'repsonse' silently disables response schema on POST /ffmpeg-settings

`server/src/api/ffmpegSettingsApi.ts:128` spells the Fastify schema key `repsonse` instead of `response`:

```ts
schema: {
  tags: ['Settings'],
  body: z.object({
    ffmpegPath: z.string(),
  }),
  repsonse: {            // <- typo
    200: FfmpegSettingsSchema,
    500: z.literal('error'),
  },
},
```

Fastify ignores unknown schema keys, so the route has no response validation or serialization at all, and the endpoint is missing its response types in the generated OpenAPI spec (and therefore in the generated web client).

One-character fix, but it should be followed by regenerating the spec and client:
```
cd server && pnpm generate-openapi
cd web && pnpm generate-client
```

Worth adding a lint rule or a schema-key allowlist so this class of typo fails the build — there are 33 route files and nothing currently catches a misspelled schema key.

oiahoon:
Closes #2029

## Summary
- correct the FFmpeg reset route response schema key
- return the settings with the same writable response shape used by the GET and PUT routes
- regenerate the versioned OpenAPI document and web client types

## Verification
- pnpm lint-changed
- pnpm turbo typecheck --filter=@tunarr/server
- pnpm turbo build --filter=@tunarr/web
- pnpm turbo test -- run
- pnpm --filter @tunarr/web extract-messages
- git diff --check


chrisbenincasa:
nice find! thanks!

PR #2036, opened 2026-08-27
fix(server): skip Jellyfin people with empty or missing names

### LINKED ISSUE #2028: Jellyfin person mapping is missing the empty-name guard that Emby has

`getJellyfinItemPersonMap` and `getEmbyItemPersonMap` are copy-pasted from one another (Jellyfin's API is a fork of Emby's, so the payload field names are identical). Emby guards against people with no name; Jellyfin does not.

Emby — `server/src/external/emby/EmbyApiClient.ts:1917-1919`:
```ts
case 'actor':
  mapping[key] = seq.collect(people, (person, idx) =>
    isNonEmptyString(person.Name)
      ? ({ name: person.Name, ... } satisfies Actor)
      : null,
  );
```

Jellyfin — `server/src/external/jellyfin/JellyfinApiClient.ts:1895-1897`:
```ts
case 'actor':
  mapping[key] = people.map(
    (person, idx) =>
      ({ name: person.Name, ... } satisfies Actor),
  );
```

So a Jellyfin item whose `People[]` contains an entry with a missing or empty `Name` produces a credit with `name: undefined`, which is then persisted. The same divergence repeats for the `writer` and `director` cases in both files.

Fix is to port Emby's `seq.collect` + `isNonEmptyString` form to the Jellyfin mapper for all three cases.

Broader context: these two mapping blocks are ~900 lines each and near-identical (`jellyfinArtworkProjection` at JF `:1859-1877` and `embyArtworkProjection` at Emby `:1881-1899` are byte-identical apart from parameter names). There are no tests over either — no `JellyfinApiClient.test.ts` or `EmbyApiClient.test.ts` exists, and the only client test, `PlexApiClient.test.ts`, asserts nothing about mapped field values. Drift like this will keep accumulating until they share an implementation.

MrStewood:
Fixes #2028

## Summary
- Mirror Emby's empty-name guard in Jellyfin person mapping for actor, writer, and director
- Drop credits with empty, whitespace, or missing `Name` instead of persisting them
- Export `getJellyfinItemPersonMap` for testing and add unit coverage

## Test plan
- [x] `pnpm test src/external/jellyfin/JellyfinApiClient.test.ts` (server)
- [x] `pnpm typecheck` (server)
- [ ] Reviewer: confirm actor/writer/director paths all use the Emby-style filter

### DIFF
--- a/server/src/external/jellyfin/JellyfinApiClient.test.ts
+++ b/server/src/external/jellyfin/JellyfinApiClient.test.ts
@@ -1,0 +1,76 @@
+import type { JellyfinItem as ApiJellyfinItem } from '@tunarr/types/jellyfin';
+import { describe, expect, it } from 'vitest';
+import { getJellyfinItemPersonMap } from './JellyfinApiClient.js';
+
+const mediaSourceUrl = 'http://jellyfin.example';
+
+type JellyfinPersonInput = NonNullable<ApiJellyfinItem['People']>[number];
+
+function person(
+  partial: Partial<JellyfinPersonInput> & Pick<JellyfinPersonInput, 'Id'>,
+): JellyfinPersonInput {
+  return partial as JellyfinPersonInput;
+}
+
+describe('getJellyfinItemPersonMap', () => {
+  it('omits people with empty, whitespace, or missing names', () => {
+    const mapping = getJellyfinItemPersonMap(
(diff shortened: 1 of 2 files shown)


chrisbenincasa:
Thanks!

PR #2050, opened 2026-09-01
refactor(server): remove dead midRollUtil module

### LINKED ISSUE #2030: Dead code: midRollUtil.ts has no production importer and its qualifier is a no-op

`server/src/services/scheduling/midRollUtil.ts` (72 lines) has **no production importer**. The only file that imports it repo-wide is its own test, `midRollUtil.test.ts:4-6`. Deleting it changes nothing.

The live implementation is `server/src/services/scheduling/midRollBreakRules.ts`, imported at `slotSchedulerUtil.ts:56`. The two implement genuinely different algorithms, so this is stale code rather than a duplicate to merge:

- `midRollUtil.calculateMidRollBreaks` (`:10`) supports only `intervalMs` plus a fixed `breakDurationMs`, computes `breakCount = ceil(dur/interval) - 1`, caps by `maxBreaks` and a slot-duration budget, and returns interval-sized **segments**.
- `midRollBreakRules.resolveBreakPoints` (`:28-90`) supports three rule types — `fixed_interval`, `percentage`, `initial_then_interval` — returns a list of **offsets** filtered against `tailBufferMs`, and has `resolveBreakDuration` for randomized `breakDurationMinMs`/`breakDurationMaxMs` ranges, which `midRollUtil` has no concept of.

Additionally, `midRollUtil.programQualifiesForMidRoll` (`:58-72`) is a no-op — its own comment concedes it cannot check subtype, and it returns `true` for every content program regardless of `config.programTypes`. So the dead code is also incorrect, which makes it a trap for anyone who revives it.

Suggested: delete `midRollUtil.ts` and `midRollUtil.test.ts`, leaving `midRollBreakRules.ts` as the single break-point authority.

andyst-dev:
Fixes #2030

Removes the dead `midRollUtil` module and its test.

- **No production importer.** `midRollUtil.ts` is referenced only by its own test, `midRollUtil.test.ts`. Replaced `grep` of `calculateMidRollBreaks` / `programQualifiesForMidRoll` / `midRollUtil` across `server/src` confirms nothing else uses it.
- **The live implementation is `midRollBreakRules.ts`** (imported at `slotSchedulerUtil.ts:56`), and the two implement genuinely different algorithms — `midRollBreakRules` supports `fixed_interval` / `percentage` / `initial_then_interval` rule types, a tail buffer, and randomized `breakDurationMinMs`/`breakDurationMaxMs` ranges that `midRollUtil` has no concept of. So this is stale code, not a duplicate to merge.
- **`programQualifiesForMidRoll` is a no-op** returning `true` for every content program regardless of `config.programTypes` — deleting it removes a trap for anyone reviving the module.

Net diff: **-72 lines + the test**, no behavior change (nothing in production imports these symbols).

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Cleanup / dead-code removal
- [ ] New feature
- [ ] Breaking change

## Testing

Verified via `grep` that `midRollUtil` (and its two exported functions) have zero references outside the module + its own test before removing.


chrisbenincasa:
Upstream break is fixed - please rebase and push to retry presubmits

andyst-dev:
Rebased onto current `main` (`100d14a5`, which includes the pnpm-overrides fix) and force-pushed to retry presubmits. The branch now contains a single commit on `a2bbc728` (only the two `midRollUtil` deletions). Build / lint / commitlint workflows have re-queued for approval — happy to adjust if anything is flagged.

PR #2051, opened 2026-09-01
ci: add the CalVer release workflow and drop the dead release tooling

chrisbenincasa:
Adds the CalVer release workflow, comments on shipped PRs and issues, and removes the old release tooling. `v1.3.15` (2026-09-15) was the last semantic-release release.

Background and rationale are in `.claude/plans/versioning-repair-plan.md` (Phase 1).

## Why now

`workflow_dispatch` workflows are only dispatchable once the file is on the default branch. Nothing about this flow can be exercised, not even a dry run, until it lands on `main`.

## What changes

- **Adds `.github/workflows/release.yml`.** Computes `YYYY.M.PATCH` from the UTC date plus the count of that month's existing stable tags. Prereleases are excluded from the stable count; `dev` produces `-dev.N` hanging off the next stable. Refuses to reuse an existing tag. Notes come from `conventional-changelog@8` in the same grouped shape semantic-release produced, and live only in the GitHub release from here on. The computed version and previous tag are passed explicitly, so notes carry the CalVer tag rather than the stale `package.json` version. Stable notes cover everything since the last stable tag; dev prerelease notes cover changes since the nearest tag of either kind. Manual, and `dry_run` defaults to `true`.
- **Adds `.github/workflows/release-comment.yml` and `.github/scripts/release-comment.sh`.** When a stable release publishes, comments on every merged PR since the previous stable tag and every issue those PRs close.
  - Closing keywords in commit messages count. `Refs #N` does not.
  - Prereleases are skipped.
  - A hidden per-tag marker makes reruns safe.
  - `workflow_dispatch` takes a tag and a `dry_run` flag.
- **Removes** `release-it.yml`, `release-please.yml`, `semantic-release.yml`, `.release-it.json`, `release.config.mjs` and `release/`.
- **Drops** seven unused release dev dependencies (`release-it`, `release-it-pnpm`, `@release-it/bumper`, `@release-it/conventional-changelog`, `semantic-release`, `@semantic-release/changelog`, `should-semantic-release`) and the `should-semantic-release` script.
- **Deletes `CHANGELOG.md`.** It stopped at `1.1.3`, and release notes now live only in GitHub Releases. Also removes its `Dockerfile` `COPY` and `.prettierignore` entry. Nothing read it at runtime.

## Version format

`YYYY.M.PATCH`, unpadded. `2026.9.0`, never `2026.09.0`. Unpadded is still valid semver, so these keep working untouched:

- `docker/metadata-action`'s `type=semver,pattern={{version}}` in `build-and-push-docker.yml`
- the `v[0-9]+.[0-9]+.[0-9]+*` Docker tag trigger
- the three-component regex and `semver.rcompare` sort in `scripts/generate-docs-script.ts`

`v1.3.15` → `v2026.9.0` increases under semver comparison, so Docker `latest`, tag ordering and the docs dropdown stay correct. `node-calver` was rejected because it renders the first release of a month as `2026.9`, which breaks all three.

## Tokens

| Workflow | Token | Reason |
|---|---|---|
| `release.yml` | `RELEASE_PLEASE_TOKEN` | Tags and releases made with `GITHUB_TOKEN` do not trigger the Docker and binary builds |
| `release-comment.yml` | `GITHUB_TOKEN` | `RELEASE_PLEASE_TOKEN` lacks comment access. Every semantic-release comment on `v1.3.15` failed with "Not allowed to add a comment to the issue/PR" |

## Verification

- Version computation checked against a fake tag set: prereleases excluded from the stable patch, an empty month starting at `.0`, dev prereleases off the next stable, and `v2026.7.9` → `2026.7.10`.
- Comment script dry-run locally against `v1.3.15` found 36 targets, identical to the 36 semantic-release tried and failed to comment on. `v1.4.0-dev.1` is skipped.
- shellcheck and actionlint pass on the new files.

Not yet run in CI. After merge, dispatch **Release (CalVer)** on `main` with `dry_run: true`. Expect `v2026.9.0`, notes since `v1.3.15`.

## Follow-ups (not in this PR)

- Phase 2: derive the OpenAPI spec filename from the release version, not `package.json`.
- Phase 4: document the versioning and rollback contract.


chrisbenincasa:
Tested this workflow against the current `dev` tree (checkout with tags `v1.3.13`, `v1.3.14`, `v1.4.0-dev.1`, `v1.2.0-dev.1..8`; root `package.json` version `1.2.0-dev.1`). Two findings, one of which I think belongs in this PR.

**Dev pre-release support itself works.** The `GITHUB_REF_NAME = dev` branch computes `YYYY.M.PATCH-dev.N` hanging off the next stable, the month scoping and prerelease exclusion from the stable count are correct, and after a stable cut the dev counter correctly moves to `vYYYY.M.(P+1)-dev.0`. Integration with the existing workflows checks out:

- `build-and-push-docker.yml`'s trigger `v[0-9]+.[0-9]+.[0-9]+*` matches `v2026.9.0-dev.1`, and the `dev` floating Docker tag fires via `contains(github.ref_name, '-dev')`.
- `build-and-release-binary.yml` attaches assets on the tag push / `release: created` event.
- Unpadded CalVer remains monotonic under semver comparison, so `latest` resolution and tag ordering stay correct.

**Finding 1 — the notes step titles the release with the stale `package.json` version.** I ran the workflow's exact command in a dev-branch checkout:

```
npx -y conventional-changelog-cli -p conventionalcommits -o RELEASE_NOTES.md -r 1
```

It produced:

```
## [1.2.0-dev.1](https://github.com/chrisbenincasa/tunarr/compare/v1.3.14...v1.2.0-dev.1) (2026-09-13)
```

The release is titled with the root `package.json` version (`1.2.0-dev.1` on dev, would be `1.3.14` on main), not the computed CalVer tag, and the compare link points backwards. The commit list itself happened to come out as `v1.3.14..HEAD` (all 26 unreleased commits), but apparently by fallback rather than by design: `conventional-changelog-cli` seeds `context.version` from `package.json`, and the range endpoint is not the computed tag. `conventional-changelog-cli@5` has no `--current-tag` flag to override this. So the first stable release would ship notes headed `1.3.14` with a bogus compare URL.

Fix options, both small:

1. Tag first (`git tag` + push via the PAT), seed the root `package.json` version with `steps.version.outputs.version`, then generate notes (the range then ends at an existing tag, and the header/compare link are correct), then `gh release create` against the existing tag.
2. Drop the CLI and drive `conventional-changelog-core` directly with explicit `context.version` and `gitRawCommitsOpts { from: prevTag, to: 'HEAD' }`, keeping the current "no tag until the release is created" flow. The previous tag can be computed from the same `git tag` listing the version step already does.

**Finding 2 — intentional, keep deferred:** the OpenAPI/docs artifact naming still derives from the stale `package.json` version (Phase 2). Fine to leave out of this PR, but it must land before the first real (non-dry-run) CalVer release, stable or dev prerelease.

**Minor, worth stating intent:** as observed, the notes range's previous tag is the last *stable* tag (`v1.3.14`, not `v1.4.0-dev.1`), so a stable cut made after dev prereleases exist will re-include the dev prerelease commits in its notes. Say whether that's the desired behavior.

Will push a fix for Finding 1 to this branch separately.

chrisbenincasa:
Pushed a fix for Finding 1 in 7183b683.

**Finding 1.** Went with a variant of option 2 that keeps the CLI: `conventional-changelog@8.1.3` (the successor to `conventional-changelog-cli@5`) accepts `--context`, `--from` and `--to`. The step now passes the computed `version`, `currentTag` and `previousTag` as context and an explicit `--from <prevTag> --to HEAD` range, so no tag is created before `gh release create` and dry runs stay side-effect free. It also needs `-i /dev/null`, since v8 otherwise prepends the old `CHANGELOG.md` to the output.

Ran the edited step locally:

- Stable, on this branch: `## [2026.9.0](…/compare/v1.3.13...v2026.9.0)`, 18 entries for the 18 `feat`/`fix`/`perf` commits since `v1.3.13`.
- Dev prerelease, in a clone with a fake `v2026.9.0-dev.0` tag: `## [2026.9.0-dev.1](…/compare/v2026.9.0-dev.0...v2026.9.0-dev.1)`, containing only the two fixes after that tag.

**Minor, intent.** Now explicit and commented in the workflow:

- Stable releases list everything since the last *stable* tag reachable from HEAD, so they deliberately roll up commits already shipped in dev prereleases. `main` users never saw those prereleases.
- Dev prereleases list only what changed since the nearest tag of either kind.

**Finding 2.** Agreed, stays in Phase 2 and must land before the first non-dry-run release.

PR #2053, opened 2026-09-01
feat(db): snapshot before migrating and refuse to run on a newer schema

chrisbenincasa:
Rollback after a bad upgrade is the behaviour users actually have, and nothing supported it. This adds a snapshot before every migration and a guard against running an old binary against a newer database.

Phase 3 of `.claude/plans/versioning-repair-plan.md`. Independent of the CalVer work in #2051 — no version number was ever going to fix this, and this is the part that actually makes a rollback possible.

## State before this change

- `DrizzleMigrator.ts` contained **zero** references to backup. Migrations ran in place against the live database.
- There was **no downgrade guard anywhere**. Nothing detected that the schema had been written by a newer Tunarr than the running binary.
- The backup machinery already existed and simply was not wired to migration: `SqliteDatabaseBackup.ts`, `ArchiveDatabaseBackup.ts`, `DatabaseBackupStrategy.ts`, and `BackupTask` bound in `TasksModule.ts`.

Both changes are small because the pieces were already there.

## Snapshot before migrating

`migrateExistingDatabase` copies the database before applying anything, using the same `SqliteDatabaseBackup` that `DatabaseCopyMigrator` already uses. **If the snapshot cannot be written, migration does not proceed.**

The file is named `db-pre-migration-<epoch>.bak` rather than reusing the existing `db-<epoch>.bak` name, and `bootstrap.ts` now rotates the two pools separately (keep newest 3 of each). Sharing one pool meant an upgrade carrying three or more full-copy migrations would evict the snapshot taken *before* any of them — precisely the one worth keeping.

Rollback becomes a one-line support answer: restore this file.

## Refuse to start on a newer schema

`DrizzleMigrator.getUnknownAppliedMigrations()` reports migration names recorded in the database that this build does not know. Non-empty means the database came from a newer Tunarr, so `migrateExistingDatabase` throws `DatabaseSchemaTooNewError`, which names the offending migration and points at the snapshot to restore.

The check runs **before** the pending-migration check. A database from the future has nothing pending and would otherwise sail straight through.

Legacy mikro-orm migrations are fast-forwarded into the migrations table under their new names, which are the provider's own keys, so they do not read as unknown here.

## Tests

Six tests in `server/src/db/DBAccess.test.ts`, all red-green proven:

- refuses a database carrying migrations this build does not know
- names the offending migration so the message is actionable
- stays out of the way of an up-to-date database
- takes no snapshot when there is nothing to migrate
- keeps the newest three of each backup pool
- snapshots the database before applying a pending migration

## Scope

5 files, +243/-3. Targets `main`. Rebased off `dev`, since under CalVer only large features go to `dev`.


chrisbenincasa:
Docs missing. docs/configure/system/backup.md documents backup/restore but says nothing about pre-migration snapshots, their 3-file rotation, or the "database created by a newer version" error. The PR body frames rollback as a "one-line support answer" — that answer deserves a short section in this doc (CLAUDE.md also asks for doc updates on behavior changes).

PR #2058, opened 2026-09-04
fix(ffmpeg): apply configured deinterlace and scaling filters to offline sessions

### LINKED ISSUE #2024: Offline session drops software deinterlace and scaling filters

`FfmpegStreamFactory.createOfflineSession` builds its `FfmpegState` without the software deinterlace and scaling filters. Both lines are commented out at `server/src/ffmpeg/FfmpegStreamFactory.ts:941-942`:

```ts
// softwareDeinterlaceFilter: this.transcodeConfig.deinterlaceFilter,
// softwareScalingAlgorithm: this.transcodeConfig.scalingAlgorithm,
```

`createErrorSession` — the structurally near-identical session builder ~90 lines above — does pass them, at `:853-854`:

```ts
softwareDeinterlaceFilter: this.ffmpegSettings.deinterlaceFilter,
softwareScalingAlgorithm: this.ffmpegSettings.scalingAlgorithm,
```

The likely reason they were commented out: they reference `this.transcodeConfig.*`, and `deinterlaceFilter` / `scalingAlgorithm` do not exist on the transcode config. Repo-wide, the only definitions are on `ffmpegSettings` (`:432`, `:435`, `:853`, `:854`).

So the fix is probably to uncomment using `this.ffmpegSettings.*` to match `createErrorSession`, rather than to delete the lines. Worth confirming that offline/flex output is intended to honour the configured scaling algorithm before applying — if it deliberately should not, the lines should be removed with a comment saying so.

Context: `createErrorSession` (`:847-877`) and `createOfflineSession` (`:935-964`) differ by only three hunks, this being one of them; the other two are a `scaledSize ??` fallback and `infiniteLoop: true`. They are copies that have drifted.

andyst-dev:
Fixes #2024

`createOfflineSession` built its `FfmpegState` with the `softwareDeinterlaceFilter` and `softwareScalingAlgorithm` lines **commented out**, referencing the non-existent `this.transcodeConfig.*`. So the offline placeholder stream ignored the configured deinterlace filter and scaling algorithm, silently falling back to the `FfmpegState` class defaults (`yadif=1` / `fast_bilinear`).

`createErrorSession` — the structurally near-identical builder sitting ~90 lines above — already passes both from `this.ffmpegSettings`. This uncomments the two lines and points them at the same `ffmpegSettings` source so both placeholder sessions behave consistently.

Regression test: `createPlaceholderSession (offline)` now asserts the captured `FfmpegState` carries the configured filters rather than the class defaults. Verified RED before the fix, GREEN after.

- [x] `vitest run src/ffmpeg/FfmpegStreamFactory.test.ts` passes (24)


chrisbenincasa:
Changes here should also have full integration tests, see https://github.com/andyst-dev/tunarr/blob/8e2998f566ac926d6866bdeee52b5099ba72a1fe/server/src/ffmpeg/builder/pipeline/software/SoftwarePipeline.local.test.ts

chrisbenincasa (changes requested):
Added comment above without formerly requesting changes, just doing it now.

andyst-dev:
Thanks for the review @chrisbenincasa. I added the requested full integration tests in `SoftwarePipeline.local.test.ts` (real ffmpeg boot), covering both filters wired in `createOfflineSession`:

- **scale from 1080p to 720p uses the configured software scaling algorithm** — asserts `flags=lanczos` lands in the actual command arguments AND probes the real output resolves to 1280x720.
- **software transcode with deinterlace=true applies the configured deinterlace filter** — asserts `yadif=0` is in the command (`softwareDeinterlaceFilter` replaces the state default) and ffmpeg exits 0 with a valid video stream.

Verified locally: SoftwarePipeline integration 7/7 (incl. the 2 new), and FfmpegStreamFactory unit 24/24 under node@22.

PR #2060, opened 2026-09-04
fix(scanner): honor pathFilter in other-video, music-video and music-artist scans

### LINKED ISSUE #2027: OtherVideo, MusicVideo and MusicArtist library scanners ignore pathFilter

`ScanContext.pathFilter` (`server/src/services/scanner/MediaSourceScanner.ts:25,39`) narrows a scan to a single item. It is honoured by:

- `MediaSourceMovieLibraryScanner.ts` — skips non-matching items at `:140-141`, and guards the mark-missing block with `if (isEmpty(context.pathFilter))` at `:162`
- `MediaSourceTvShowLibraryScanner.ts` — same pattern at `:161`, `:202`, `:397`, `:617`
- all five `Local*` scanners

It is **not referenced at all** by:

- `MediaSourceOtherVideoScanner.ts`
- `MediaSourceMusicVideoScanner.ts`
- `MediaSourceMusicArtistScanner.ts`

So a targeted single-item scan on an other-video, music-video or music-artist library walks and re-processes the entire library instead of the one requested item. Filtering is applied client-side in the loop (not pushed into the API query), so items are still all seen and the mark-missing diff stays correct — the impact is wasted work and scan time, not incorrect state. Worth confirming that holds for every `getVideos` implementation before closing.

These files are near-verbatim copies of one another — `diff` of the extracted `scanInternal` bodies of `MediaSourceOtherVideoScanner` and `MediaSourceMusicVideoScanner` is a **single line**, `ProgramType.OtherVideo` vs `ProgramType.MusicVideo`. `pathFilter` support is a feature the copies never picked up.

One inconsistency in the other direction, worth fixing at the same time: `MediaSourceOtherVideoScanner.scanInternal` wraps its body in `try/finally` so `mediaSourceProgressService.scanEnded` always runs (`:109`, `:151-153`). `MediaSourceMovieLibraryScanner` does not, so an exception mid-scan leaves the progress indicator hung.

andyst-dev:
Fixes #2027

The OtherVideo, MusicVideo and MusicArtist library scanners never referenced `ScanContext.pathFilter`, so a targeted single-item scan re-walked the entire library instead of narrowing to the one requested item. The Movie/TvShow scanners already honor it — they skip non-matching items and guard the mark-missing pass with `if (isEmpty(context.pathFilter))`.

This ports that pattern to the three scanners:
- `MediaSourceOtherVideoScanner` / `MediaSourceMusicVideoScanner` (near-identical linear scans): skip a video whose `externalId` doesn't match `pathFilter`, and skip the whole mark-missing pass when a filter is set.
- `MediaSourceMusicArtistScanner` (artist → album → track walk): skip non-matching artists and guard the mark-missing pass the same way.

The scan narrows to the requested item, so a targeted scan no longer reprocesses the whole library (the mark-missing diff stays correct because it's suppressed when a filter is active).

Regression test (`MediaSourceOtherVideoScanner.test.ts`): with `pathFilter` set, only the matching item is processed and nothing is marked missing; without it, every item is scanned and the unseen persisted item is marked missing. Verified RED before the fix, GREEN after.

- [x] `vitest run src/services/scanner/MediaSourceOtherVideoScanner.test.ts` passes
- [x] prettier clean

### DIFF
--- a/server/src/services/scanner/MediaSourceMusicArtistScanner.ts
+++ b/server/src/services/scanner/MediaSourceMusicArtistScanner.ts
@@ -5,7 +5,14 @@ import type {
   Season,
   Show,
 } from '@tunarr/types';
-import { differenceWith, flatten, head, round, values } from 'lodash-es';
+import {
+  differenceWith,
+  flatten,
+  head,
+  isEmpty,
+  round,
+  values,
+} from 'lodash-es';
 import type { GetProgramGroupingById } from '../../commands/GetProgramGroupingById.ts';
 import type { ProgramGroupingMinter } from '../../db/converters/ProgramGroupingMinter.ts';
 import type { ProgramDaoMinter } from '../../db/converters/ProgramMinter.ts';
@@ -152,7 +159,7 @@ export abstract class MediaSourceMusicArtistScanner<
   ): Promise<void> {
     this.mediaSourceProgressService.scanStarted(context.library.uuid);
 
-    const { library } = context;
+    const { library, pathFilter } = context;
     const existingArtists =
       await this.programDB.getExistingProgramGroupingDetails(
         library.uuid,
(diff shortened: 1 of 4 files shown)


chrisbenincasa:
LGTM

PR #2062, opened 2026-09-04
fix(tasks): don't pause on-demand channels with zero sessions

### LINKED ISSUE #2025: OnDemandChannelStateTask pauses channels that have no sessions on every tick

`server/src/tasks/OnDemandChannelStateTask.ts:44-52`:

```ts
const allChannelSessions = this.sessionManager.getAllSessionsForChannel(
  channel.uuid,
);
if (
  every(allChannelSessions, (session) => session.numConnections() === 0)
) {
  this.onDemandService.pauseChannel(channel.uuid, stopTime).catch(...)
}
```

lodash `every([])` returns `true`, so a channel with **zero** sessions satisfies the condition and `pauseChannel` is called on every tick. The loop also runs over every channel, not just on-demand ones — the `// TODO filter down to on-demand only...` at `:41` is unresolved.

Two further inconsistencies in the same area, worth resolving together:

1. This task counts raw connections (`session.numConnections() === 0`). `SessionManager.pauseChannelIfNecessary` (`:436-494`) counts only *external* connections — it filters out Tunarr's own at `:468`:
   ```ts
   (conn) => !(conn.userAgent?.includes('Tunarr') ?? false),
   ```
   So the two paths disagree about whether a channel is idle.

2. `api/streamApi.ts:177-179` resumes inline via `onDemandChannelService.resumeChannel(...)`, bypassing `SessionManager.resumeChannelIfNecessary` (`:425`).

There is effectively no test coverage here: `stream/SessionManager.test.ts` constructs the manager with `new (SessionManager as any)(...)` (`:131`) and covers only session-replacement races and staleness; `services/OnDemandChannelService.test.ts` is 35 lines asserting DI singleton scope. Pause/resume and the cursor math have no tests.

andyst-dev:
Fixes #2025

`OnDemandChannelStateTask` calls `pauseChannel` when every session for a channel has zero connections. With **no** sessions at all, lodash `every([])` returns `true`, so a channel with zero sessions (and, per the unresolved `TODO`, non-on-demand channels too) got `pauseChannel` invoked on **every tick** — idle work on the pause/scheduler path.

The fix guards the condition on `allChannelSessions.length > 0`, so a channel with no sessions is left alone. Two further inconsistencies the issue lists (raw vs external connection counting, and the resume path bypass) are deliberately out of scope here — they're noted as follow-ups in the issue.

Regression test (`OnDemandChannelStateTask.test.ts`): zero sessions → `pauseChannel` not called (RED before the fix, GREEN after); all-idle sessions → paused; an active session → not paused.
- [x] `vitest run src/tasks/OnDemandChannelStateTask.test.ts` passes (3)
- [x] `pnpm lint-changed --branch` clean
- [x] prettier clean

### DIFF
--- a/server/src/tasks/OnDemandChannelStateTask.test.ts
+++ b/server/src/tasks/OnDemandChannelStateTask.test.ts
@@ -1,0 +1,77 @@
+import type { IChannelDB } from '@/db/interfaces/IChannelDB.ts';
+import type { OnDemandChannelService } from '@/services/OnDemandChannelService.js';
+import type { SessionManager } from '@/stream/SessionManager.js';
+import { describe, expect, test, vi } from 'vitest';
+
+// OnDemandChannelStateTask imports OnDemandChannelService (and transitively
+// SessionManager) through Inversify bindings that only resolve inside a booted
+// container. Mock the DI classes before importing the task, mirroring
+// SessionManager.test.ts.
+vi.mock('@/services/OnDemandChannelService.js', () => ({
+  OnDemandChannelService: class {},
+}));
+vi.mock('@/services/EventService.js', () => ({
+  EventService: class {},
+}));
+
+import { OnDemandChannelStateTask } from './OnDemandChannelStateTask.ts';
+
+function makeSut(sessionConnections: number[]) {
+  const channelDB = {
+    loadAllLineupConfigs: vi.fn().mockResolvedValue({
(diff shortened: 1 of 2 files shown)


chrisbenincasa (approved):

chrisbenincasa:
Thanks!

PR #2064, opened 2026-09-06
fix(hls): remove _minByIp entry by connection IP (defensive)

### LINKED ISSUE #2045: HLS playlist window stops advancing for everyone once a client falls behind (or leaves) – _minByIp never gets cleaned up
---

### What happens

On a channel with more than one viewer, the HLS playlist served by Tunarr can stop moving. Every request for `stream.m3u8` returns the same 20 segments, players drain their buffer and stall, even though ffmpeg keeps writing new segments to disk every few seconds.

It's enough that one client stops requesting segments – a paused player, a tab in the background, a phone that went to sleep – and all the other viewers of that channel freeze with it. Worse, once a client has watched the channel and gone away (closed the tab, switched channel from another device), the window stays pinned to the last segment it asked for, for as long as the session lives.

### Why (1.3.13, still the same on `main` as far as I can see)

`BaseHlsSession.ts` keeps `_minByIp = new Map<string, number>()`, keyed by client IP, and `minSegmentRequested` returns the minimum of the values. `HlsSession.trimPlaylist()` passes that as `segmentNumber` to `HlsPlaylistMutator`, which serves `maxSegmentsToKeep` segments starting from `minSegmentRequested - 10`. So the slowest client ever seen decides the window for everybody.

The cleanup is where it actually breaks:

```ts
override removeConnection(token: string) {
  ...
  this._minByIp.delete(token);
}
```

The map is keyed by `clientIp` (`this._minByIp.set(clientIp, parsed)`), but `removeConnection` deletes by `token`. That never matches, so entries are never removed. A client that left keeps holding the window back until the session is torn down.

### How I reproduced it

- Start a channel, watch it in a browser through a reverse proxy (so the IPs are distinct).
- From a second machine, request a couple of segments and then stop (or just pause the player).
- Keep polling `stream.m3u8` from the first client: the segment list stays identical (in my case sn 1707–1726 for 76 s) while `ls` on the segment directory shows new files every 4 s. hls.js runs out of buffer and stalls.
- Simulated "forgotten" clients (`X-Forwarded-For: 10.7.7.7` with `segment 0` requested once) pin the window permanently.

### What I'm running as a workaround

I patched the binary so that `minSegmentRequested` always returns a huge number; the filter comes back empty and the mutator falls through to `takeRight(allSegments, maxSegmentsToKeep)`, i.e. always the latest 20 segments. With that, a channel with 10+ concurrent viewers (some paused, some on bad Wi-Fi) has been stable for days. A client that falls more than ~80 s behind gets a 404 on an old segment and re-syncs to the live edge, which is fine for live TV.

### Suggested fix

- Key the map by the same thing `removeConnection` receives (the connection token), or look the IP up from the token before deleting.
- Consider not using the minimum across clients at all: a slow client should only affect its own playback. Serving the last N segments (with a generous N) and letting late clients 404 and re-sync is what most live HLS origins do.
- If the per-client window is kept, entries should also expire after some time without requests, independent of `removeConnection`.

Happy to test a fix against the setup above.

andyst-dev:
Fixes #2045

You're right — thanks for the careful trace. The root cause I cited (the `req.query.token ?? v4()` handshake at `streamApi.ts#L140`) feeds `getOrCreateConcatSession`, i.e. the **MPEG-TS** route, not HLS. Every `BaseHlsSession` connection is registered under the client IP (`req.ip`) on `main`: master playlist handshake (`getOrCreateHlsSession`/`getOrCreateHlsSlowerSession`) and the fragment route both pass `req.ip` as the token. So `token === ip` on every production HLS path, and the existing `_minByIp.delete(token)` already reaches the right entry. My regression test is correct but the framing as a root-cause fix for #2045 was wrong.

I've re-scoped this PR as **defensive hardening only**:

- `removeConnection` now deletes `_minByIp` by the connection's IP rather than the token, with a fallback to the token for unknown connections. Today that's behavior-neutral (`ip === token`); it locks in the invariant that a non-IP-keyed `BaseHlsSession` connection can never leak its entry and pin the playlist window for other viewers.
- Reworded the inline comment to not cite the MPEG-TS handshake.
- Added a fallback-branch test (unknown token → delete by token is a no-op that leaves live viewers untouched).

Regarding your suspects for the actual #2045 freeze:

1. Reproduced: I did **not** reproduce the freeze on `main`; I wrote the test purely from the code path. I can't currently claim `main` freezes.
2. Token ≠ IP on HLS: I cannot point to any path on `main` where `BaseHlsSession` gets a non-IP token — which is exactly why this is hardening, not the root-cause fix.
3. `trustProxy`: worth ruling out on the reporter's setup, agreed.
4. **The playlist-po11 / heartbeat path is the strong lead.** On `streamApi.ts#L306` the fragment route records a heartbeat for *every* file request, while `onSegmentRequested` (L327) is skipped for `stream.m3u8` (L308-324 returns early). So a client that only polls `stream.m3u8` stays "alive" forever while its last `_minByIp` entry (from its final segment request) pins the window — untouched by this PR. That matches the report's "paused player / backgrounded tab" symptom.

I'd be happy to chase that lead as a follow-up: e.g. either record the heartbeat only for segment requests (not playlist polls), or have a playlist-only poll refresh the connection's `_minByIp` to the current live edge so it stops pinning. Want me to open that as a separate PR?

Test count: `vitest run src/stream` — 10 tests in `BaseHlsSession.test.ts` pass (8 existing + 2 in this PR); full `src/stream` 94 pass. `pnpm lint-changed` clean.

### DIFF
--- a/server/src/stream/hls/BaseHlsSession.test.ts
+++ b/server/src/stream/hls/BaseHlsSession.test.ts
@@ -87,6 +87,24 @@ describe('BaseHlsSession', () => {
       expect(session.minSegment).toBe(10);
     });
 
+    it('removes the _minByIp entry when the start-handshake token differs from the client IP', () => {
+      // Bug #2045: the fragment route keys _minByIp by client IP, but the
+      // connection created by the start handshake is keyed by a UUID token,
+      // so removeConnection(token) never matched the IP entry — a departed
+      // client kept pinning the playlist window for every other viewer.
+      const startToken = '0f2c1e0a-8e6f-4b0b-9d2a-3c9f7a1b2c3d';
+      session.addConnection(startToken, makeConnection('192.168.1.1'));
+      session.addConnection('192.168.1.2', makeConnection('192.168.1.2'));
+      session.onSegmentRequested('192.168.1.1', 'data000010.ts');
+      session.onSegmentRequested('192.168.1.2', 'data000050.ts');
+
+      // Client 1 departs; its connection is removed under the start token
+      session.removeConnection(startToken);
+
+      expect(session.minByIp.has('192.168.1.1')).toBe(false);
+      expect(session.minSegment).toBe(50);
+    });
+
     it('removes stale IP entries from _minByIp after removeStaleConnections', () => {
       session.addConnection('192.168.1.1', makeConnection('192.168.1.1'));
       session.addConnection('192.168.1.2', makeConnection('192.168.1.2'));
--- a/server/src/stream/hls/BaseHlsSession.test.ts
+++ b/server/src/stream/hls/BaseHlsSession.test.ts
@@ -88,10 +88,10 @@ describe('BaseHlsSession', () => {
     });
 
     it('removes the _minByIp entry when the start-handshake token differs from the client IP', () => {
-      // Bug #2045: the fragment route keys _minByIp by client IP, but the
-      // connection created by the start handshake is keyed by a UUID token,
-      // so removeConnection(token) never matched the IP entry — a departed
-      // client kept pinning the playlist window for every other viewer.
+      // Defensive invariant: _minByIp is keyed by client IP; if any caller
+      // ever registers a connection under a non-IP token, removeConnection
+      // must still release the IP entry (or the departed client pins the
+      // playlist window for every other viewer).
       const startToken = '0f2c1e0a-8e6f-4b0b-9d2a-3c9f7a1b2c3d';
       session.addConnection(startToken, makeConnection('192.168.1.1'));
       session.addConnection('192.168.1.2', makeConnection('192.168.1.2'));
@@ -103,6 +103,18 @@ describe('BaseHlsSession', () => {
 
       expect(session.minByIp.has('192.168.1.1')).toBe(false);
       expect(session.minSegment).toBe(50);
+    });
+
+    it('falls back to deleting by token for an unknown connection (no-op)', () => {
+      session.addConnection('192.168.1.1', makeConnection('192.168.1.1'));
+      session.onSegmentRequested('192.168.1.1', 'data000010.ts');
+
+      // An unknown token is not a registered connection; removing it must
+      // not disturb the live viewer's entry.
+      session.removeConnection('192.168.99.99');
+
+      expect(session.minByIp.has('192.168.1.1')).toBe(true);
+      expect(session.minSegment).toBe(10);
     });
 
     it('removes stale IP entries from _minByIp after removeStaleConnections', () => {
--- a/server/src/stream/hls/BaseHlsSession.ts
+++ b/server/src/stream/hls/BaseHlsSession.ts
@@ -99,13 +99,14 @@ export abstract class BaseHlsSession<
     const connection = this.connections()[token];
     super.removeConnection(token);
 
-    // _minByIp is keyed by client IP (see onSegmentRequested), but the
-    // connection token can differ from the IP: the start handshake keys
-    // connections by a UUID (streamApi `req.query.token ?? v4()`). Deleting
-    // by token never matched, so a departed client's entry kept pinning the
-    // playlist window to its last-requested segment for every other viewer.
-    // Remove by the connection's IP (falling back to the token when the
-    // connection is unknown) so the entry is released on disconnect.
+    // _minByIp is keyed by client IP (see onSegmentRequested). Today every
+    // BaseHlsSession connection is registered under the client IP, so the
+    // token and the IP coincide and deleting by either reaches the entry.
+    // Delete by the connection's IP instead of the token so the behavior
+    // holds even if a caller ever keys a connection under a non-IP token
+    // (the entry would otherwise be leaked and keep pinning the playlist
+    // window for other viewers). Fall back to the token when the connection
+    // is unknown to preserve the current no-op behavior for stale tokens.
     if (connection?.ip) {
       this._minByIp.delete(connection.ip);
     } else {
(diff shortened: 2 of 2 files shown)


chrisbenincasa:
Thanks for the PR! The fix itself looks good — read-before-delete ordering is correct,
the fallback preserves current behavior for unknown tokens, and the regression test is
well targeted (94/94 in `src/stream` on my checkout, typecheck and lint clean).

However, I traced the connection-token flow and I don't think the root cause as written
can trigger on current `main` for HLS sessions. Details:

### Where the UUID token actually lives

The cited handshake (`req.query.token ?? v4()`) is in the **MPEG-TS route**
[`streamApi.ts#L140`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/streamApi.ts#L140),
which feeds `getOrCreateConcatSession`
[`streamApi.ts#L143`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/streamApi.ts#L143).
That session is a `ConcatSession` (`DirectStreamSession` → `Session`) — it has **no
`_minByIp`** and uses the base `Session.removeConnection`. So a UUID token can never
interact with the HLS playlist-window map.

### Every `BaseHlsSession` connection is registered under the client IP

- Master playlist handshake:
  [`streamApi.ts#L400`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/streamApi.ts#L400)
  (`hls`) and
  [`streamApi.ts#L437`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/streamApi.ts#L437)
  (`hls_slower`) both pass `req.ip` as the token, which lands in
  [`SessionManager.ts#L391`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/stream/SessionManager.ts#L391)
  `session.addConnection(token, connection)`.
- Fragment route:
  [`streamApi.ts#L301`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/streamApi.ts#L301)
  registers `session.addConnection(req.ip, …)` and
  [`streamApi.ts#L327`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/streamApi.ts#L327)
  keys `_minByIp` by `req.ip`.

So for HLS sessions `token === ip` on every production path, and the existing
`_minByIp.delete(token)` already removes the right entry. I confirmed this was also true
in v1.3.13. Stale cleanup does run:
[`Session.removeStaleConnections`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/stream/Session.ts#L288)
is invoked every minute by
[`CleanupSessionsTask`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/services/startup/ScheduleJobsStartupTask.ts#L51),
so a departed client's entry is released within ~2–3 min on `main`.

In other words: this PR is safe defensive hardening, but as written it may not be the
fix for the freeze in #2045. Could you confirm a few things on your end?

1. **Did you reproduce #2045 on a current `main` build, or only on 1.3.13?** If the
   freeze reproduces on `main`, the token/IP mismatch can't be the cause there — could
   you share the repro (config, proxy setup, timings)?
2. **Can you point to any code path where a `BaseHlsSession` subclass receives
   `addConnection` with a token ≠ client IP?** I couldn't find one on `main`.
3. **What is `req.ip` behind your reverse proxy?** If `trustProxy` is off, all clients
   collapse to the proxy's IP and `_minByIp` becomes a single shared entry — worth
   ruling out.
4. If it still repros: **did the departed client keep polling `stream.m3u8`?** Note the
   fragment route records a heartbeat for *every* file request
   ([`streamApi.ts#L306`](https://github.com/chrisbenincasa/tunarr/blob/main/server/src/api/streamApi.ts#L306))
   but skips `onSegmentRequested` for playlist requests — a playlist-only client stays
   "alive" forever while its stale `_minByIp` entry pins the window. That path is
   untouched by this PR and is my main suspect if the freeze is real on `main`.

### Suggested test

I'd love to see a route-level regression test that locks the invariant through the
actual `streamApi` routes: inject requests with distinct `remoteAddress` values,
assert the handshake registers the connection under the client IP (never a UUID), and
assert stale cleanup releases the window for the surviving viewer. I prototyped this —
it passes on both `main` and this branch (i.e. it documents current behavior and would
fail if anyone later switches the HLS handshake to UUID tokens).

### Nits

- The inline comment cites the `.ts` (MPEG-TS) handshake as if it were the HLS flow —
  could you reword it to describe the fallback without that citation?
- A tiny test for the fallback branch (unknown token → delete by token) would be nice.

If you re-verify and find `main` still freezes, I'd suggest we dig into the
playlist-poll/heartbeat path together — happy to help reproduce.

andyst-dev:
Thanks for the detailed trace — you're correct. The `req.query.token ?? v4()` handshake is the MPEG-TS `ConcatSession` route, not HLS; every `BaseHlsSession` connection is keyed by `req.ip`, so `token === ip` on all HLS paths and the old `_minByIp.delete(token)` already worked. I've re-scoped this PR to defensive hardening (IP-keyed delete + fallback test, no claim of fixing #2045) and updated the body to spell out your playlist-poll lead (L306 heartbeat vs L327 `onSegmentRequested` skipped for `stream.m3u8`) as the real suspect. I'm happy to chase that as a separate PR — say the word.

andyst-dev:
Closing this PR.

As you demonstrated in your review, `token === ip` on every `BaseHlsSession` path on `main` (the `req.query.token ?? v4()` handshake I cited is the MPEG-TS `ConcatSession` route, not HLS), so this change is behavior-neutral today. It doesn't fix the #2045 freeze, and I'd rather not leave a zero-effect PR in the queue.

Thanks for the trace — it's a better lead than my fix was. I'm going to pursue your playlist-poll hypothesis as the real fix instead: the fragment route records a heartbeat for every file request (L306) but skips `onSegmentRequested` for `stream.m3u8` (L308-324 early-return), so a playlist-only client stays alive forever while its stale `_minByIp` entry pins the window. I'll validate it with a route-level regression test reproducing the freeze before opening anything.

Apologies for the false-start root cause.

PR #2069, opened 2026-09-08
fix(scheduling): apply per-slot pad override to every program in the slot

chrisbenincasa:
## Problem

A time slot with a pad-time override only had that override applied to the **first** program in the slot. Every subsequent program in the same slot, and the flex distributor, fell back to the schedule-level `padMs`.

The user-visible symptom is that the slot *after* an overridden slot silently disappears from the schedule.

Reported setup:
- Schedule-level pad `:00,:30` (30m), max lateness 5m
- 6:00 slot: Bluey (7m episodes), pad override to 5m marks
- 6:30 slot: Little Bear, no override

Observed: Bluey respects its 5m override, Little Bear is never scheduled at all.

## Cause

In `scheduleTimeSlots`:

- `TimeSlotService.ts:272` used `currSlot.padMs ?? schedule.padMs` — correct
- `TimeSlotService.ts:296` (the loop filling the rest of the slot) used `schedule.padMs`
- `TimeSlotService.ts:344` (`distributeFlex`) used `schedule.padMs`

So Bluey ep1 was padded to 10m (5m grid) but ep2 was padded to 30m (30m grid). The slot ran 40m inside a 30m window, the cursor landed at 6:40, the top-of-loop realignment to the global 30m grid pushed it to 7:00, and the 6:30 slot was then 30m late. That is over the 5m max lateness, so it was flexed — and being the last slot of the day, the flex ran until 6:00 the next morning.

## Fix

Hoist the effective pad into a single `slotPadMs` and use it at all three sites.

This is not a recent regression. It dates to 57b7f63a ("feat: add ability to set slot padding per-slot in time slot editor", #1614), which introduced the override but only patched one of the three call sites.

Random slots are unaffected: per-slot `padMs` only exists on `BaseTimeSlot` (`types/src/api/TimeSlots.ts:22`) and `RandomSlotsService` never reads it.

## Follow-up cleanup

`RandomSlotsService` does **not** have this bug — random slots have no per-slot pad override (`BaseRandomSlotSchema` carries no `padMs`), and its `distributeFlex` call is correctly gated to `padStyle === 'episode'`.

It did, however, have the same duplication that allowed the time slot drift: `handleFixedDurationSlot` inlined `padStyle === 'slot' ? 1 : padMs` for the first program while every subsequent program went through the `createPaddedProgram` helper that computes the identical expression. Both resolved to the same value, so the second commit is a no-op refactor that collapses the inline into the helper and drops the now-unused local destructure.

## Test plan

- [x] New regression test `per-slot pad override does not starve the following slot` in `TimeSlotService.test.ts`, which asserts both slots produce content, that the overridden slot stays on its 5m grid and inside its 30m window, and that the un-overridden slot keeps the 30m grid
- [x] Verified the new test **fails** on the pre-fix code (`expected 0 to be greater than 0` — no programs from the second slot) and passes after
- [x] `pnpm vitest run src/services/scheduling/` — 114/114 passing
- [x] `pnpm turbo typecheck` clean

Resulting schedule after the fix:

```
06:00  bluey-ep1      7m
06:07  flex           3m
06:10  bluey-ep2      7m
06:17  flex           3m
06:20  bluey-ep3      7m
06:27  flex           3m
06:30  littlebear-ep1 25m
```


Wolfie713:
Is this related to the issue I reported (Discord)?

PR #2073, opened 2026-09-10
fix(plex): invalidate canonical items when Plex labels change

### LINKED ISSUE #2021: Invalidate Plex items when Labels change - should cause scan to occur
Fix the PlexMediaCanonicalizer to incorporate a check for Labels, which get saved and indexed today, but never cause existing items to change their hash

andyst-dev:
## Summary
The Plex content-hash canonicalizers (movie, show, season, episode, artist, album, track) folded `key`/`title`/`addedAt`/`updatedAt` and media/tag fields into the sha1, but never `Label`. Editing an item's labels in Plex updates `updatedAt`? — no: in practice a pure label edit doesn't surface as a content change here, so the canonical id stays identical and the item is never re-canonicalized (stale pick/modification). This makes label changes participate in the hash so they invalidate items (#2021).

## Changes
- New private `hashPlexTags(hash, key, tags)` helper.
- Call it with `'label'` for `Label` on all seven item canonicalizers.

## Test
`PlexMediaCanonicalizers.test.ts` — added "changes id when a label is added" / "changes id when a label tag changes" for movie, show and episode. 134 tests pass, no tsc errors.

Fixes #2021

### DIFF
--- a/server/src/services/PlexMediaCanonicalizers.test.ts
+++ b/server/src/services/PlexMediaCanonicalizers.test.ts
@@ -212,6 +212,25 @@ describe('PlexMediaCanonicalizer - movie', () => {
     );
   });
 
+  it('changes id when a label is added', () => {
+    expect(canonicalizer.getCanonicalId(baseMovie)).not.toBe(
+      canonicalizer.getCanonicalId({
+        ...baseMovie,
+        Label: [{ tag: 'Favorites' }],
+      }),
+    );
+  });
+
+  it('changes id when a label tag changes', () => {
+    const labeled = { ...baseMovie, Label: [{ tag: 'Favorites' }] };
+    expect(canonicalizer.getCanonicalId(labeled)).not.toBe(
+      canonicalizer.getCanonicalId({
+        ...labeled,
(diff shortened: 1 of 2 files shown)


chrisbenincasa:
Thanks again - I'm going to add an AI policy to the repo soon. As such, can you ensure that changes that are AI-generated, even partially, have proper attribution? Thanks!

chrisbenincasa:
Can a greater refactoring happen here? PlexMedia is the root structurally of all of these? It must share a set of common fields? Perhaps we can have a common method that applies canonicalization at the PlexMedia level to each, including Labels, and then do the per-type specific stuff?

andyst-dev:
Hey Chris, fair enough on the attribution, I will add it myself whenever something is AI-assisted.

And yeah, the per-type hashing is repetitive. PlexMedia is the common base, so we could hash the shared fields once and just add the type-specific stuff after. Happy to redo it that way if you want.

chrisbenincasa:
We can keep this change focused for now and revisit later. Don't want to blow up the scope too much.

PR #2074, opened 2026-09-10
fix(server): normalize stored language codes to ISO 639-2/T (#2044)

### LINKED ISSUE #2044: Language codes are persisted unnormalized, so the DB and search index hold a mix of ISO 639-2/B and /T
Found while fixing #1960. Not user-reported directly, but it is the systemic root of that bug and it degrades search in a way users would experience as "the filter doesn't work".

## Summary

ISO 639-2 assigns 20 languages two three-letter codes — a bibliographic (`/B`) code from the English name and a terminological (`/T`) code from the native name. German is both `ger` and `deu`; French `fre`/`fra`; Dutch `dut`/`nld`; also Chinese, Czech, Greek, Icelandic, Macedonian, Malay, Persian, Romanian, Slovak, Albanian, Armenian, Basque, Burmese, Georgian, Maori, Tibetan, Welsh.

**Tunarr normalizes on one ingest path and not the others**, so `program_media_stream.language` and `program_subtitles.language` — and the Meilisearch index built from them — contain a mix of code sets that varies by media source.

## Evidence

Writers that persist the provider's code verbatim:

- `server/src/external/plex/PlexApiClient.ts:2116` (video), `:2156` (audio), `:2196` (subtitles) — `languageCodeISO6392: audioStream.languageCode`
- `server/src/external/jellyfin/JellyfinApiClient.ts:1238`, `:1258` — `languageCodeISO6392: nullToUndefined(audioStream.Language)`
- `server/src/external/emby/EmbyApiClient.ts:1292`, `:1310` — same pattern
- `server/src/db/converters/ProgramMinter.ts:185` writes it to `program_media_stream.language`; `:309` writes embedded subtitles to `program_subtitles.language`

The counter-example that creates the mixture:

- `server/src/stream/FfprobeStreamDetails.ts:113`, `:141` normalize with `LanguageService.getAlpha3TCode`, so the local-file path stores `/T`
- `server/src/db/converters/ProgramMinter.ts:327` writes sidecar subtitles from `LocalSubtitlesService`, also normalized to `/T`

Net effect: **the same program can hold two code sets** — embedded subtitles unnormalized from the provider, sidecar subtitles normalized to `/T`.

There is also no validation of language codes anywhere in `server/src/api/`, so any string a client sends is persisted verbatim. `server/src/db/schema/SubtitlePreferences.ts:13-14` comments the column as "iso6392 - 3-letter code" without specifying which set.

## Consequences

**1. Search filtering silently returns wrong results.** `MeilisearchService.getUniqueStreamLanguages` (`:1669-1685`) copies `languageCodeISO6392` into the index verbatim, `audioLanguages`/`subtitleLanguages` are filterable facets (`:170-171`), and filter values pass through unnormalized into the Meili expression (`:1596`, `:1604`). There is no language-aware `SearchFilterValueMutator` — only `MediaSourceNameSearchMutator` and `LibraryNameSearchMutator` exist in `server/src/services/search/`.

So `audio_language = deu` matches local-file programs and misses Jellyfin ones **in the same query**. The DSL exposes `audio_language` / `subtitle_language` as free text (`shared/src/util/searchUtil.ts:62-63`, `:409-410`) with no normalizer, and there is no faceted autocomplete constraining the value (`web/src/helpers/searchBuilderConstants.ts` has no language entry), so users type these by hand.

**2. Facet counts split.** The facet endpoint (`server/src/api/programmingApi.ts:172-199`) lists `ger` and `deu` as two separate values with two separate counts.

**3. Saved smart collections silently lose items.** `server/src/db/SmartCollectionsDB.ts` persists these filters, so a collection built when a source reported one code stops matching when it reports the other.

**4. Churn.** `server/src/db/program/ProgramMetadataRepository.ts:338` compares `existing.language !== incoming.language` to decide subtitle re-extraction, and `server/src/services/LocalMediaCanonicalizer.ts:226` folds the raw code into the media-item content hash. A `ger`↔`deu` flip triggers redundant re-extraction and re-canonicalization. Benign, but wasteful.

## Why #2041 / #2042 don't fix this

Those normalize at **comparison** time, in memory, on the streaming path. That is the right fix for stream selection and it makes playback correct regardless of stored code set. It does nothing for stored data or the search index, which is where the remaining damage is.

## Proposed fix

1. **Normalize on write.** Route the provider clients through `LanguageService.normalizeToAlpha3T` before persisting, keeping the raw value only if it cannot be resolved. Decide whether to normalize in the API clients or centrally in `ProgramMinter` — central is fewer sites, but the clients are where the provider-specific quirks live.
2. **Backfill existing rows.** `server/src/tasks/fixers/` already has the pattern (`Fixer` in `fixer.ts`, run at startup by `FixerRunner.ts:28-55`, must be idempotent — `BackfillProgramExternalIds.ts` is a close analogue).
3. **Reindex.** Normalized rows only reach search after reindexing; `MeilisearchService.ts:1469` (`deleteAllDocuments`) plus the existing index build is the mechanism.
4. **Optionally normalize query values too**, as defense in depth for indexes built before the backfill — a language `SearchFilterValueMutator` alongside the existing two.

## Decisions needed

- Normalize in the API clients or centrally in `ProgramMinter`?
- Is a startup fixer the right home for a backfill that touches every stream row, or should it be an explicit maintenance task?
- Should the backfill force a reindex, or wait for the next natural scan?
- Do we normalize `program_subtitles.language` too, given `:309` stores `'unknown'` as a sentinel?

## Related

- #1960 — the user-reported symptom
- #2026 — divergent subtitle matching between transcode and remux paths
- #2041, #2042 — the comparison-time fixes

## Also found in the same sweep (separate, smaller)

- `SubtitleStreamPicker` skips the language check entirely when `languageCodeISO6392` is absent, so an untagged stream matches any requested language. Second half of #2026.
- `web/src/components/channel_config/ChannelSubtitlePreferencesTable.tsx:137` passes `values={[]}` and never dedupes, so any language can be added to a channel twice.
- `languageCodeISO6391` is declared in `server/src/stream/types.ts:54,71` and read in five places as a `??` fallback, but **no code in the server ever assigns it**. Those branches are dead.

andyst-dev:
Fixes #2044

## Problem
`program_media_stream.language` / `program_subtitles.language` — and the Meilisearch facets built from them — hold a **mix of ISO 639-2 /B and /T codes** depending on which ingest path wrote each program (German `ger` vs `deu`, French `fre` vs `fra`, …). A language filter matches only the half of the library whose source stored one set.

The root cause is two-fold:

1. **`LanguageService.getAlpha3TCode` never converts /B → /T.** Its /B branch called `alpha3TToAlpha2(input)`, which the `@cospired/i18n-iso-languages` lib leaves `undefined` for a /B code — so every `ger` passed through unchanged. The library exposes `alpha3BToAlpha2` for exactly this conversion.
2. **`ProgramMinter` wrote the provider's code verbatim** (`stream.languageCodeISO6392`, subtitles `languageCodeISO6392`, sidecar subtitle language) on the external-ingest path, bypassing normalization entirely.

## Fix
- **`LanguageService`**: switch the /B branch to `alpha3BToAlpha2` → `alpha2ToAlpha3T`, so `getAlpha3TCode`/`normalizeToAlpha3T` now correctly resolve `/B → /T`. This is the true systemic fix — it also makes the local-file path (`FfprobeStreamDetails` → `getAlpha3TCode`) correct for /B inputs.
- **`ProgramMinter`**: routes all three language write sites through `LanguageService.normalizeLanguageCode` (a thin wrapper over `normalizeToAlpha3T` that keeps the raw value when it cannot be resolved, including the `'unknown'` subtitle sentinel). Every future program persists its language as ISO 639-2/T.
- **Backfill** (`NormalizeLanguageCodesFixer`, registered in `FixerModule`, `canRunInBackground`, **idempotent** — only rows whose code actually resolves to a different /T code are updated): repairs the existing `program_media_stream` and `program_subtitles` rows on the next startup, **paged** (one distinct input code at a time, `WHERE language = ? LIMIT 500`) so it neither materializes a whole table nor approaches SQLite's bound-variable limit.
- **Search index healing**: after each page the fixer republishes the language facets of the programs it touched (`MeilisearchService.updatePrograms`, partial update of `audioLanguages` / `subtitleLanguages` only), so the facets and saved filters work right after the upgrade instead of waiting for a rescan. Only documents already present in the index are touched, so an unscanned program never gets a stub document. Fixers run before Meilisearch is ready, hence the new `MeilisearchService.waitUntilReady()`.
- `SubtitlePreferences.languageCode` comment now states the canonical set.

## Validation
- `ProgramMinter.test.ts` (new): `/B → /T` (`ger→deu`, `fre→fra`, `dut→nld`, `chi→zho`), already-/T unchanged, 2-letter → /T, English name → /T, unresolvable + `'unknown'` kept verbatim, absent passthrough.
- `NormalizeLanguageCodesFixer.test.ts` (new, real sqlite): pages the backfill across batches, waits for search readiness before touching anything, republishes the normalized facets, and skips programs that are not in the index. **22/22 pass.**
- `tsc --noEmit` and `oxlint` clean.
- Manual run on a scratch instance (sqlite + Meilisearch, no library rescan, no reindex): DB `ger,fre,dut` with `audioLanguages = deu` returning 0 hits → after one restart DB `deu,fra,nld` with `deu` → 2 hits, `ger` → 0, `fra`/`fre` subtitle facets correct, and a second restart changes nothing (idempotent).

## Notes / scope left
- Reindex: no longer deferred. The fixer heals the index itself, per touched program, at startup. It deliberately does not trigger a full reindex; happy to move this refresh into a standalone task if you would rather keep fixers DB-only.
- I did **not** add the optional language `SearchFilterValueMutator` (defense-in-depth for pre-backfill indexes / hand-typed query values) — flag if you want it included as a follow-up here.
- The related `SubtitleStreamPicker` language-check and `languageCodeISO6391` dead-assignment items from the issue are separate and intentionally out of scope.

### DIFF
--- a/server/src/db/converters/ProgramMinter.test.ts
+++ b/server/src/db/converters/ProgramMinter.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from 'vitest';
+import { normalizeLanguageCode } from './ProgramMinter.js';
+
+// The ISO 639-2 /B→/T normalization that keeps stored language codes in one
+// canonical set. Regression coverage for #2044: a provider's bibliographic
+// code must turn into the terminological code Tunarr stores elsewhere
+// (FfprobeStreamDetails / LocalSubtitlesService already write /T), and an
+// unresolvable value must be kept as-is rather than dropped.
+describe('normalizeLanguageCode (#2044)', () => {
+  it.each([
+    // /B → /T (the reported class: German 'ger' vs 'deu')
+    ['ger', 'deu'],
+    ['fre', 'fra'],
+    ['dut', 'nld'],
+    ['chi', 'zho'],
+  ])('maps ISO 639-2/B "%s" to its /T equivalent "%s"', (input, expected) => {
+    expect(normalizeLanguageCode(input)).toBe(expected);
+  });
+
+  it.each(['deu', 'fra', 'nld', 'por', 'jpn', 'eng'])(
+    'keeps an already-/T code "%s" unchanged',
+    (input) => {
+      expect(normalizeLanguageCode(input)).toBe(input);
+    },
+  );
+
+  it.each(['de', 'fr', 'nl', 'pt', 'ja'])(
+    'maps a 2-letter code "%s" to its /T code',
+    (input) => {
+      const out = normalizeLanguageCode(input);
+      expect(out).toBeDefined();
+      expect(out).toHaveLength(3);
+    },
+  );
+
+  it('maps an English language name to its /T code', () => {
+    expect(normalizeLanguageCode('japanese')).toBe('jpn');
+    expect(normalizeLanguageCode('German')).toBe('deu');
+  });
+
+  it.each(['unknown', 'xxyy', 'abc123'])(
+    'keeps an unresolvable value "%s" verbatim (incl. the subtitle sentinel)',
+    (input) => {
+      expect(normalizeLanguageCode(input)).toBe(input);
+    },
+  );
+
+  it('passes through an empty or absent value', () => {
+    expect(normalizeLanguageCode(undefined)).toBeUndefined();
+    expect(normalizeLanguageCode('')).toBe('');
+  });
+});
--- a/server/src/db/converters/ProgramMinter.test.ts
+++ b/server/src/db/converters/ProgramMinter.test.ts
@@ -49,4 +49,11 @@ describe('normalizeLanguageCode (#2044)', () => {
     expect(normalizeLanguageCode(undefined)).toBeUndefined();
     expect(normalizeLanguageCode('')).toBe('');
   });
+
+  it('coerces a null code to undefined', () => {
+    // Plex/Emby subtitle language fields are nullable; the function must not
+    // blow up and must return undefined so callers can fall back to their
+    // default (e.g. ?? 'unknown').
+    expect(normalizeLanguageCode(null)).toBeUndefined();
+  });
 });
--- a/server/src/db/converters/ProgramMinter.ts
+++ b/server/src/db/converters/ProgramMinter.ts
@@ -19,6 +19,7 @@ import { injectable } from 'inversify';
 import { compact, find } from 'lodash-es';
 import { match, P } from 'ts-pattern';
 import { v4 } from 'uuid';
+import { LanguageService } from '@/services/LanguageService.js';
 import {
   HasMediaSourceInfo,
   MediaSourceMovie,
@@ -47,6 +48,21 @@ import {
   NewProgramWithRelations,
 } from '../schema/derivedTypes.js';
 import { CommonDaoMinter } from './CommonDaoMinter.ts';
+
+/**
+ * Normalize a stored language code to ISO 639-2/T, keeping the raw value when
+ * it can't be resolved (e.g. an unknown provider code, or the 'unknown'
+ * sentinel on subtitles). Without this, ingest paths that write a provider's
+ * code verbatim mix ISO 639-2 /B ("ger") and /T ("deu") in the database and the
+ * search index built from it — so a language filter matches half the library
+ * depending on which source a program came from (#2044). Plain 2-letter and
+ * already-/T codes pass through unchanged.
+ */
+export function normalizeLanguageCode(
+  code: string | undefined,
+): string | undefined {
+  return code ? (LanguageService.normalizeToAlpha3T(code) ?? code) : code;
+}
 
 /**
  * Generates Program DB entities for Plex media
@@ -182,7 +198,7 @@ export class ProgramDaoMinter {
           colorPrimaries: stream.colorPrimaries ?? null,
           default: stream.default ?? false,
           //TODO: forced: stream.forced
-          language: stream.languageCodeISO6392,
+          language: normalizeLanguageCode(stream.languageCodeISO6392),
           pixelFormat: stream.pixelFormat,
           title: stream.title,
         } satisfies NewProgramMediaStream;
@@ -306,7 +322,7 @@ export class ProgramDaoMinter {
         programId,
         createdAt: now,
         updatedAt: now, // Do we need to use mtime?
-        language: subtitle.languageCodeISO6392 ?? 'unknown',
+        language: normalizeLanguageCode(subtitle.languageCodeISO6392) ?? 'unknown',
         subtitleType:
           subtitle.streamType === 'subtitles' ? 'embedded' : 'sidecar',
         default: subtitle.default ?? false,
@@ -324,7 +340,7 @@ export class ProgramDaoMinter {
         codec: subtitle.codec,
         createdAt: now,
         updatedAt: now, // Do we need to use mtime?
-        language: subtitle.language,
+        language: normalizeLanguageCode(subtitle.language),
         subtitleType: subtitle.subtitleType,
         default: subtitle.default ?? false,
         forced: subtitle.forced ?? false,
--- a/server/src/db/converters/ProgramMinter.ts
+++ b/server/src/db/converters/ProgramMinter.ts
@@ -59,9 +59,9 @@ import { CommonDaoMinter } from './CommonDaoMinter.ts';
  * already-/T codes pass through unchanged.
  */
 export function normalizeLanguageCode(
-  code: string | undefined,
+  code: string | null | undefined,
 ): string | undefined {
-  return code ? (LanguageService.normalizeToAlpha3T(code) ?? code) : code;
+  return code ? (LanguageService.normalizeToAlpha3T(code) ?? code) : code ?? undefined;
 }
 
 /**
@@ -340,7 +340,7 @@ export class ProgramDaoMinter {
         codec: subtitle.codec,
         createdAt: now,
         updatedAt: now, // Do we need to use mtime?
-        language: normalizeLanguageCode(subtitle.language),
+        language: normalizeLanguageCode(subtitle.language) ?? 'unknown',
         subtitleType: subtitle.subtitleType,
         default: subtitle.default ?? false,
         forced: subtitle.forced ?? false,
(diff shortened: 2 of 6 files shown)


andyst-dev:
Fixed the build that was failing. The typecheck kept complaining because subtitle language codes can be null in Plex, but the normalize function only expected string or undefined. Three call sites broke for that reason.

So: normalizeLanguageCode now handles null gracefully (empty strings still pass through as-is), and the extra subtitle language falls back to unknown the same way the embedded one already did. Added a small regression test for the null case.

Branch is updated, typecheck is clean and the converter tests pass 21/21.

chrisbenincasa (changes requested):
**Verdict: Request changes.** The diagnosis and the `LanguageService` one-line fix are correct and verified (21/21 new tests pass, typecheck + `lint-changed --branch` clean; library behavior confirmed: `alpha3TToAlpha2('ger')` → `undefined`, `alpha3BToAlpha2('ger')` → `de` → `deu`). Minter write-path coverage is complete — I found no DB write site for `program_media_stream.language` / `program_subtitles.language` that bypasses it. Two blockers before this lands:

**1. (Blocking) The fixer doesn't heal the search index, so the user-facing bug survives the fix.** Nothing rebuilds Meilisearch at startup — `deleteAllDocuments()` is only reachable from the manually-scheduled dangling-programs task, and per-program index updates only happen on scan/upsert. After this backfill, the `audioLanguages`/`subtitleLanguages` facets and saved filters keep matching the old mixed codes until every program is rescanned or a full manual reindex happens — i.e. the symptom in the issue title ("filter doesn't work") persists post-upgrade. The issue's proposed fix listed reindex as step 3; the PR defers it. Please wire a reindex (or an equivalent per-program search-document refresh of the touched rows) after the backfill, or justify why not.

**2. (Blocking) Fixer doesn't scale to large libraries.** It selects *all* rows of both tables into memory, then issues one `UPDATE ... WHERE uuid IN (...)` per normalized code. A single group (e.g. every `ger` stream) can exceed SQLite's 32766 bound-variable limit, and the full-table materialization grows unboundedly with library size. Suggest grouping by the *original* code and updating via `WHERE language = <orig>` — one statement per distinct input value, no uuid lists, no variable-limit risk. (Or page the scan like `BackfillProgramExternalIds` does.)

**3. (Minor)** `canRunInBackground: true` is misleading — `runInternal` is fully synchronous (better-sqlite3 sync, no awaits), so it blocks the event loop during startup and never yields; either yield between updates or drop the flag.

**4. (Nit)** Both new files lack a trailing newline (`pnpm fmt` will fix); `normalizeLanguageCode` might live better on `LanguageService` beside `normalizeToAlpha3T`. The fixer has no unit test — consistent with the other fixers, so optional.

chrisbenincasa:
I've merged a lot - mind resolving the merge conflicts here?

andyst-dev:
Pushed `d20c8da4`, which addresses all four points.

**1. Index healing (blocking).** The backfill now republishes the search documents of the rows it touched. Per batch it recomputes the affected programs' `audioLanguages` / `subtitleLanguages` from the DB and calls `searchService.updatePrograms([{ id, audioLanguages, subtitleLanguages }])`.

Design choices:
- It is a partial update, not a full document rebuild, so the fields the scan path owns (grouping parents, tags, state, externalIds) cannot be clobbered by a backfill.
- Only programs already present in the index are updated (`getPrograms` first). A program that was never scanned does not get a stub document; it gets indexed with the corrected code on its next scan.
- Facet values go through the same normalization helper as the write path, so a page published mid-backfill already carries its final value, and unresolvable values (the `'unknown'` subtitle sentinel) are kept verbatim.
- Fixers run before Meilisearch is ready, so the fixer awaits a new `MeilisearchService.waitUntilReady()`, resolved at the end of `start()`.

Verified end to end on a real instance (SQLite + Meilisearch, no library rescan, no index rebuild):
- before: DB `ger,fre,dut`; filter `audioLanguages = deu` → 0 hits, `= ger` → 2 hits
- after one restart: log `Normalized 6 stored language code(s) to ISO 639-2/T and refreshed their search documents (media streams: 5, subtitles: 1)`; DB `deu,fra,nld`; `deu` → 2 hits, `ger` → 0, `subtitleLanguages = fra` → 1, `= fre` → 0
- second restart: nothing re-normalized and the facets are unchanged (idempotent)

**2. Scaling (blocking).** Every statement is paged now. The fixer walks one distinct input code at a time with `WHERE language = ? LIMIT 500`, updates those rows, refreshes their documents, then loops until a short batch comes back. No table is materialized in memory, and no statement can approach SQLite's 32766 bound-variable limit (500 uuids per update, search refresh chunked at 100 ids).

I kept the page loop rather than a single `UPDATE ... WHERE language = ?` so the refresh stays incremental: each page publishes the documents it changed instead of buffering an entire library.

**3. `canRunInBackground` (minor).** Now accurate: `runInternal` awaits the readiness gate and the search updates, so it yields to the event loop instead of blocking startup.

**4. Nits.** Prettier and trailing newlines fixed; `normalizeLanguageCode` moved onto `LanguageService` beside `normalizeToAlpha3T`, with `ProgramMinter` and the fixer both calling it; the fixer now has a unit test covering the paging, the readiness gate, the facet republish and the "not in the index" case.

22 tests pass, `tsc --noEmit` and `oxlint` clean. I have updated the PR description to match. Happy to move the search refresh behind a flag or into a standalone task if you would rather keep fixers DB-only.

andyst-dev:
Rebased onto `main` again (`127bb0f9`), conflicts resolved.

For the record, the conflicts you mentioned on the 20th were already resolved in `d20c8da4`, I just never said so in that reply. Main moved another 47 commits since then, so this is a fresh rebase on top of it.

Only one conflict needed real thought, in the subtitle minter: #1993 rewrote that block, so instead of keeping my own ternary for `subtitleType` it now uses the `isExternal` flag from that change, with the language normalization sitting on top of it.

22/22 tests pass and the typecheck output is identical to `main`. Whenever you get a chance to look again.

PR #2075, opened 2026-09-10
fix(external): round Jellyfin/Emby RunTimeTicks durations to integer ms (#2032)

### LINKED ISSUE #2032: Jellyfin/Emby fractional durations are stored as REAL in INTEGER columns
Found during a backend architecture review. Not user-reported, and **not reproduced** — see the caveat at the bottom before investing time.

## Root cause

Jellyfin and Emby report durations in `RunTimeTicks` (100-ns units). Ten canonicalization sites divide by `10_000` without rounding, so any `RunTimeTicks` that is not a multiple of 10,000 yields a fractional millisecond value:

- `server/src/external/jellyfin/JellyfinApiClient.ts` — `:1151`, `:1530`, `:1700`, `:1776`, `:1848`
- `server/src/external/emby/EmbyApiClient.ts` — `:1211`, `:1556`, `:1726`, `:1798`, `:1870`

All of the form:
```ts
duration: movie.RunTimeTicks / 10_000,
```

**Two sites in the same files already guard this**, which is good evidence the fractional case is real rather than theoretical — `JellyfinApiClient.ts:1295` and `EmbyApiClient.ts:1321`:
```ts
duration: +dayjs.duration(Math.ceil((source.RunTimeTicks ?? 0) / 10_000)),
```

The guard was simply never applied to the item-canonicalization paths.

From there the value flows unrounded into `Program.duration`, declared `integer().notNull()` (`server/src/db/schema/Program.ts:62`), via `ProgramMinter.ts:111/213/358/426/480/537`. `program.duration` becomes `lineupItem.durationMs`, and `sum(durationMs)` becomes `channel.duration`, also `integer().notNull()`.

Plex is unaffected — it reports integer milliseconds.

## What SQLite actually does

It does **not** throw. INTEGER affinity stores a REAL when the conversion would be lossy. Verified against the project's own better-sqlite3:

```
insert 1234.5             -> stored 1234.5             typeof real
insert 1234.0             -> stored 1234               typeof integer
insert 4358589839.408408  -> stored 4358589839.408408  typeof real
insert null               -> SQLITE_CONSTRAINT_NOTNULL
```

Drizzle's `SQLiteInteger` defines no `mapToDriverValue`, and Kysely is identity, so nothing coerces on the way down. The result is a silent schema violation: `typeof(duration) = 'real'` in a column declared INTEGER.

## Fix at the source, not at the DB layer

Change the ten sites to `Math.round(X.RunTimeTicks / 10_000)`. `Math.round` rather than the `Math.ceil` used at `:1295`/`:1321` — for a real media duration, nearest-ms recovers the true value and sub-ms inflation serves no purpose.

**Do not fix this at the channel layer.** There are 11 write sites to `channel.duration` (`BasicChannelRepository.ts:61,87,273,294,339`; `LineupRepository.ts:333,388,479,779`; plus `StreamProgramCalculator.ts:107` and `RegenerateChannelLineupCommand.ts:64,93`), and `syncChannelDuration` (`BasicChannelRepository.ts:294-311`) would actively revert any rounding:

```ts
const lineupDuration = sum(map(lineup.items, (item) => item.durationMs));
if (lineupDuration !== channel.duration) {
  // writes the raw fractional sum back
}
```

Round the channel value while leaving `durationMs` fractional and that comparison is permanently true, so the fractional value is rewritten on every sync.

Historical note: PR #1684 attempted a channel-layer `Math.round` and was closed unmerged on 2026-04-22, because `c95d8028` (repository split, 2026-04-10) deleted the lines it patched. It covered 3 of the 11 sites, so it would not have held even if merged.

## Related: the self-heal path has a gap

`ReconcileProgramDurationsTask` propagates corrected program durations into lineups, so a rescan would repair existing rows. But at `:160-163` it calls:

```ts
await this.channelDB.saveLineup(channel.uuid, {
  ...lineup,
  items: newLineupItems,
});
```

The spread carries `startTimeOffsets` through unchanged, so offsets stay stale while durations change. There is already a `// TODO` at `:135-136` acknowledging this. Worth fixing in the same change — dropping `startTimeOffsets` from the spread so they are recomputed.

## Suggested scope

1. `Math.round` at the ten client sites.
2. Drop `startTimeOffsets` from the `ReconcileProgramDurationsTask` spread.
3. Optionally a fixer in `server/src/tasks/fixers/` doing `UPDATE program SET duration = round(duration) WHERE typeof(duration) = 'real'`, so existing rows are repaired without waiting for a rescan.
4. Optionally add `.int()` to the duration fields in `types/src/schemas/programmingSchema.ts` and `channelSchema.ts` so this fails loudly at the API edge in future. Check nothing legitimately sends a fraction first.

## Caveat — worth confirming before investing

This was not reproduced. There are no Jellyfin/Emby fixtures in the repo, and the dev database is Plex-only (3432 programs, all `typeof(duration) = 'integer'`). The evidence is the unrounded division plus the two pre-existing `Math.ceil` guards around the same expression.

One query against a real Jellyfin or Emby user database settles it:

```sql
SELECT count(*) FROM program WHERE typeof(duration) = 'real';
```

andyst-dev:
Fixes #2032

## Problem
Jellyfin and Emby report durations as `RunTimeTicks` (100-ns units). Ten item-canonicalization sites divided by `10_000` **without rounding**, so any `RunTimeTicks` that is not a multiple of 10,000 produced a fractional-millisecond value. SQLite stores that as a `REAL` in `Program.duration` despite the column being declared `integer().notNull()` — a silent schema violation that then propagates into `lineupItem.durationMs` and `sum(durationMs)` → `channel.duration`. Plex is unaffected (integer ms).

The two same-file sites that already guarded this used `Math.ceil` — good evidence the fractional case is real, the guard was just never applied to the item-canonicalization paths.

## Fix
`Math.round(RunTimeTicks / 10_000)` at all **10 unrounded sites** (Jellyfin: `JellyfinApiClient.ts` movie/episode/track/2× video; Emby: `EmbyApiClient.ts` same five shapes, same order). Nearest-ms rounding recovers the true duration for real media (sub-ms inflation from `Math.ceil` serves no purpose).

`channel.duration` write sites and `syncChannelDuration` are deliberately untouched, per the issue: rounding there would be reverted by the raw-sum write-back.

## Validation
- `tsc` clean on both changed files (after building `@tunarr/types` + `@tunarr/shared`)
- Existing `JellyfinApiClient` test: **2/2 passed** (Emby has no dedicated unit test; both files are the same shape)
- `lint-changed` clean, `prettier --check` clean
- No un-rounded `duration: X.RunTimeTicks / 10_000` site remains in either file (10/10 verified)

### DIFF
--- a/server/src/external/emby/EmbyApiClient.ts
+++ b/server/src/external/emby/EmbyApiClient.ts
@@ -1221,7 +1221,7 @@ export class EmbyApiClient extends MediaSourceApiClient<EmbyItemTypes> {
       ),
       mediaSourceId: this.options.mediaSource.uuid,
       libraryId: '', // We can't know this at this point...
-      duration: movie.RunTimeTicks / 10_000,
+      duration: Math.round(movie.RunTimeTicks / 10_000),
       externalId: movie.Id,
       artwork: compact([
         this.embyArtworkProjection('poster', movie, 'Primary'),
@@ -1566,7 +1566,7 @@ export class EmbyApiClient extends MediaSourceApiClient<EmbyItemTypes> {
         episode,
         this.options.mediaSource.uuid,
       ),
-      duration: episode.RunTimeTicks / 10_000,
+      duration: Math.round(episode.RunTimeTicks / 10_000),
       artwork: compact([
         this.embyArtworkProjection('poster', episode, 'Primary'),
         this.embyArtworkProjection('banner', episode, 'Banner'),
@@ -1736,7 +1736,7 @@ export class EmbyApiClient extends MediaSourceApiClient<EmbyItemTypes> {
               }
             : null,
         ) ?? [],
-      duration: track.RunTimeTicks / 10_000,
+      duration: Math.round(track.RunTimeTicks / 10_000),
       externalId: track.Id,
       artwork: compact([
         this.embyArtworkProjection('poster', track, 'Primary'),
@@ -1808,7 +1808,7 @@ export class EmbyApiClient extends MediaSourceApiClient<EmbyItemTypes> {
       ),
       mediaSourceId: this.options.mediaSource.uuid,
       libraryId: '', // We can't know this at this point...
-      duration: video.RunTimeTicks / 10_000,
+      duration: Math.round(video.RunTimeTicks / 10_000),
       externalId: video.Id,
       artwork: compact([
         this.embyArtworkProjection('poster', video, 'Primary'),
@@ -1880,7 +1880,7 @@ export class EmbyApiClient extends MediaSourceApiClient<EmbyItemTypes> {
       ),
       mediaSourceId: this.options.mediaSource.uuid,
       libraryId: '', // We can't know this at this point...
-      duration: video.RunTimeTicks / 10_000,
+      duration: Math.round(video.RunTimeTicks / 10_000),
       externalId: video.Id,
       artwork: compact([
         this.embyArtworkProjection('poster', video, 'Primary'),
--- a/server/src/external/jellyfin/JellyfinApiClient.ts
+++ b/server/src/external/jellyfin/JellyfinApiClient.ts
@@ -1161,7 +1161,7 @@ export class JellyfinApiClient extends MediaSourceApiClient<JellyfinItemTypes> {
       ),
       mediaSourceId: this.options.mediaSource.uuid,
       libraryId: '', // We can't know this at this point...
-      duration: movie.RunTimeTicks / 10_000,
+      duration: Math.round(movie.RunTimeTicks / 10_000),
       externalId: movie.Id,
       artwork: compact([
         this.jellyfinArtworkProjection('poster', movie, 'Primary'),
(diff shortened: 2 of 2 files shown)


chrisbenincasa:
I'm not sure this is the "real" fix here tbh. A schema change is more apt, but more expensive. Even if we did go with this change, I'd encourage a round down rather than a round up, because ffmpeg can act weird when duration is even slightly longer than the streams within it. This  is a touchy change - I'm not convinced it should totally AI driven.

andyst-dev:
Fair — I agree the rounding at read time isn't the real fix and I'd rather not ship a band-aid. My take: `RunTimeTicks`/duration should be normalized once where it's mapped into the schema (a proper integer-ms duration column), not re-derived per read. That's the cleaner fix but it's a bigger change touching the DB schema and the ingest path. I'm happy to go that direction (round down / floor where a value can't be avoided), but since it reshapes the schema I wanted to confirm the approach with you before writing it.

Issue #2087, opened 2026-09-13
[Bug]: A channel with zero programs, or a schedule referencing an unresolvable program, hangs the guide-builder and takes down the entire server


augustosc-eu:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on GitHub (I've searched it) — closest related reports are #1962 and #1973, cross-referenced below; neither reports this specific symptom or the local-media-source trigger path.
- [x] I'm using an up to date version of Tunarr (1.3.14, latest at time of writing).

### Description of the bug

A channel whose schedule ends up referencing zero programs, or a program ID that the guide-builder can't resolve, sends Tunarr's guide-scheduler into an unbounded loop that pegs one CPU core at ~100% and makes the **entire HTTP server unresponsive** — not just that one channel. Because guide-building for every channel runs synchronously at startup, before the HTTP listener binds, this becomes a genuine crash-loop: restarting doesn't help, since the same channel is rebuilt (and re-loops) every time. Recovery requires manually removing the offending channel/custom-show rows from the SQLite database directly.

This looks like the same architectural class of issue already raised in **#1962** ("Plex ratingKey identity desync") and **#1973** ("startup silently deletes program rows below a specific timestamp boundary") — both describe Tunarr treating a source-local, mutable program identity as durable, and a rescan/reconciliation path that can desync or cascade-delete `program` rows out from under a channel that still references them. @chrisbenincasa's comment on #1962 explicitly flagged that "there are two (three really with local) other media sources to consider" beyond Plex — this report is the **local media source** case of that same gap, reproduced twice, with a materially worse consequence: instead of one channel failing to play (#1962's symptom, with health checks staying green), the entire server becomes unresponsive for every channel.

### Environment

- Tunarr version: **1.3.14** (macOS native app build, `tunarr-macos`)
- OS: macOS 26.5.2 (build 25F84), Apple Silicon, single-instance (no Docker)
- Media source: local, `music_videos` type, populated/republished via the HTTP API by an external app that publishes AI-scheduled channels (custom shows + a time/random schedule) against `/api/custom-shows` and `/api/channels/:id/programming`

### Steps to reproduce

**Trigger A — zero-program schedule accepted and stored:**
[...]

PR #2089, opened 2026-09-13
fix: stop scheduler hangs on empty custom shows and reject invalid schedules (#2087)

chrisbenincasa:
## Summary

Fixes the scheduling hang reported in #2087, where a slot pointing at an empty custom show made schedule generation loop forever and blocked the server.

**Scheduler (`85d839b6`)**
- Random slots skip a slot that places nothing without advancing time, and fill with flex until another slot is eligible.
- Sequential selection skips exhausted slots; cooldown waits no longer overshoot the requested window.
- Programs without a positive duration are excluded from random and time slot scheduling.
- Direct scheduler calls reject non-positive fixed durations and invalid dynamic counts with an error naming the slot.

**API (`5c43dcc2`)**
- Random slot saves and previews use a strict schema (at least one slot, positive fixed duration, whole-number program count).
- Saves and previews reject slots referencing a custom show with no schedulable content. Regenerating already-saved rules still falls back to flex.
- Manual lineups reject zero-duration items and unknown program IDs.
- The derived lineup is validated before any SQL, lineup file, cache or guide change. Failures return 400 with the reason, including from worker threads.
- `PUT /api/custom-shows/:id` replaces content atomically: omitted `programs` keeps content, `[]` clears it, unknown IDs reject the whole update.

**Web (`85da35cf`)**
- New-slot defaults and the type picker skip empty custom shows.
- A saved slot whose custom show is empty or deleted keeps its selection, shows a warning in both slot tables, and blocks Save.
- The custom-show field no longer shows a different show than the slot references.
- Preview and save errors show the server's validation message.
- Docs cover the empty-show behavior and custom-show content replacement.

## Notes for review

- **OpenAPI spec and client:** only this PR's contract changes are applied. A full regeneration currently also rewrites unrelated endpoints (request-body defaults, channel create/update), and a full client regeneration breaks `ShowSearchSlotProgrammingForm.tsx` because the committed client already lags the committed spec. That drift is left for a separate change.
- **Worker pool timeout:** with `TUNARR_USE_WORKER_POOL` enabled, a timed-out task still leaves its worker busy. That is handled in a follow-up PR.

## Test plan

- [x] `pnpm turbo typecheck`
- [x] `pnpm turbo test`, with new regressions for:
  - empty custom shows under sequential, uniform and weighted choice
  - cooldowns and linked slots
  - zero-duration content
  - invalid durations
  - custom-show replacement against real SQLite
  - endpoint rejections that leave stored lineups unchanged and the server responsive
- [x] `pnpm lint-changed`
- [ ] Manual browser check of the time and random slot editors: an empty or deleted custom show on a saved slot, and a failed save showing the server message


server/src/db/CustomShowDB.ts
```diff
@@ -125,11 +126,19 @@ export class CustomShowDB {
       return null;
     }
 
-    if (updateRequest.programs && updateRequest.programs.length > 0) {
-      await this.upsertCustomShowContent(show.uuid, updateRequest.programs);
+    // An omitted field leaves membership alone. An explicit list, including an
+    // empty one, replaces it. Resolve every ID before changing anything.
+    const replacementPrograms = updateRequest.programs;
```
chrisbenincasa: **Data loss.** Omitting `programs` from a PUT no longer means "leave membership alone". `UpdateCustomShowRequestSchema` is `CreateCustomShowRequestSchema.partial()`, and the create schema has `programs: z.array(...).default([])` — the default survives `.partial()`, so an omitted `programs` parses to `[]` and falls into this replacement branch, deleting every `custom_show_content` row. The deleted `programs && programs.length > 0` guard made that path inert on main.

The web form hits this on every save of a synced show: `EditCustomShowForm.tsx:156` sends `programs: data.syncEnabled ? [] : data.programs`, so contents are wiped until the next sync run restores them, and until then every random slot referencing the show fails the new empty-show check.

Fix: make `programs` genuinely optional in the update schema (no default), so `undefined` reaches this branch, or treat an explicit `[]` as a no-op.

server/src/services/scheduling/RandomSlotsService.ts
```diff
@@ -376,33 +429,53 @@ export class RandomSlotScheduler {
...
-          case 'show':
-          case 'custom-show':
-          case 'filler':
-          case 'smart-collection':
-            break;
+      const spec = slot.durationSpec;
+      if (spec.type === 'fixed') {
+        if (!Number.isFinite(spec.durationMs) || spec.durationMs <= 0) {
+          throw new Error(
+            `Slot ${index} (type = ${slot.type}) has fixed duration ${spec.durationMs}; it must be a positive number of milliseconds`,
```
chrisbenincasa: **Regression for existing installs.** `generateSchedule` calls `validateSchedule()` unconditionally (line 222), so these new positivity checks also run when regenerating a schedule already in the DB. Schedules saved before this branch can have a fixed `durationMs: 0` or a fractional dynamic `programCount` — the permissive on-disk `RandomSlotScheduleSchema` admits both — and those now fail every regeneration. The plain Error escapes the worker, `findBadRequestError` finds nothing, the endpoint answers 500 with no slot context, and every channel-settings save that triggers regeneration fails the same way. On main the same schedule played with flex padding.

The reference check at `SlotSchedulerHelper.ts:117-119` is deliberately skipped on regeneration so playback can fall back to flex; the positivity checks need the same treatment (validate strictly on save, repair or tolerate on regeneration).

server/src/services/TunarrWorker.ts
```diff
@@ -84,6 +85,10 @@ export class TunarrWorker {
         type: 'error',
         requestId: req.requestId,
         message: result.error.message,
+        httpCode:
```
chrisbenincasa: **This branch's 400s never fire in production.** `result.error instanceof TypedError` is always false: `Result.attemptAsync` (`types/result.ts:25-31`) funnels every throw through `WrappedError.fromError` (`types/errors.ts:7-15`), which builds an anonymous subclass with an empty message and the original error on `.cause`. So `httpCode` here (and at `:114`) is never populated and `message` is always `''`.

With the worker pool on (the default — `container.ts` reads `TUNARR_USE_WORKER_POOL` defaulting true), `TunarrWorkerPool.ts` skips the 400 branch and rejects `new Error('')`, and the caller sees a 500 with an empty body instead of the intended 400 plus reason. The branch's tests assert 400 and stay green because `server/vitest.config.ts:48` forces `TUNARR_USE_WORKER_POOL=false`.

Recover the original error before the instanceof check — e.g. unwrap `WrappedError` via its `.cause`, or make `toWrappedError` preserve the typed error as an instance.

server/src/api/channelsApi.ts
```diff
@@ -571,6 +572,18 @@ export const channelsApi: RouterPluginAsyncCallback = async (fastify) => {
           },
         };
       } else if (req.body.type === 'random') {
+        const strictSchedule = StrictRandomSlotScheduleSchema.safeParse(
```
chrisbenincasa: **The hang #2087 fixes for random slots is still persistable through time slots.** The `type === 'time'` branch above (`:559-572`) still runs only `validateSlotGroups`, which checks iteration-group consistency and nothing else, while the time preview endpoint (`:828`) has used `StrictTimeSlotScheduleSchema` since main.

A client POSTs a time schedule with a fractional `startTime` (e.g. `1234.567`): the save succeeds with 200, and on the next regeneration `scheduleTimeSlots` enters `while (timeCursor.isBefore(upperLimit))` with a slot length that truncates to zero, so `pushFlex(0.567)` leaves the cursor bit-identical and the loop never exits. Reproduced on branch code: the loop blocked the event loop and the process had to be killed at 60s; vitest's own timeout never fired. Same uninterruptible-hang class as #2087, now persistable through the API this branch hardened for random slots only.

The other hole the strict schema documents, `padMs: 0`, is caught by the new `assertValidLineup` (NaN durations rejected as 400), so the fractional-`startTime` case is the one that reaches the scheduler. Run `StrictTimeSlotScheduleSchema` here too.

types/src/api/RandomSlots.ts
```diff
@@ -200,4 +200,44 @@ export const RandomSlotScheduleSchema = z.object({
...
+
+    if (!Number.isInteger(spec.programCount) || spec.programCount <= 0) {
+      ctx.addIssue({
+        code: 'custom',
+        path: ['slots', index, 'durationSpec', 'programCount'],
+        message: `Slot ${index}: program count must be a positive whole number, got ${spec.programCount}`,
+      });
+    }
+
+    if (slot.type === 'flex' || slot.type === 'redirect') {
```
chrisbenincasa: **The editor cannot repair the state this schema rejects.** `newSlotForType` (`EditRandomSlotDialogContent.tsx:271-345`) returns no `durationSpec` for flex/redirect, so switching a Dynamic slot's Type to Flex or Redirect leaves the old dynamic spec in place through `reset((prev) => ({ ...prev, ...slot }))`, and the Fixed/Dynamic `ToggleButtonGroup` renders only when `programType !== 'flex' && programType !== 'redirect'` (`:413`).

Concrete repro: set a slot's duration to Dynamic, change its Type to Flex or Redirect, press Save. The server answers 400 "Slot N: flex slots need a fixed duration", and the dialog in that state offers no Fixed/Dynamic control nor a duration column. The schedule is unsavable until the user discovers the workaround of switching the type away, choosing Fixed, and switching back. Either clear the durationSpec on type change or show a duration control for flex/redirect.

web/src/helpers/slotSchedulerUtil.ts
```diff
@@ -269,6 +269,47 @@ export function slotOrderOptions(
...
+  customShowId: string,
+): CustomShowAvailability {
+  const option = programOptions.find(
+    (opt): opt is CustomShowProgramOption =>
+      opt.type === 'custom-show' && opt.customShowId === customShowId,
+  );
+  if (option === undefined) {
+    return 'missing';
+  }
+  return option.programCount > 0 ? 'available' : 'empty';
```
chrisbenincasa: **Web and server disagree on what "empty" means.** This counts `programCount > 0`, a membership-row count, while the server's new empty-show check counts only members with a finite positive duration. The two definitions diverge for rows whose durations are zero or non-finite.

A custom show assembled from items with 0 or non-finite durations (a local-file scan reporting 0, or legacy rows) shows as available here — no alert, Save enabled — and the API then answers 400 "references custom show X, which has no content to schedule" (or the empty 500 from the worker-pool error path above), with no pointer to which program in the show is at fault. Align this check with the server's definition of usable.

server/src/services/scheduling/SlotSchedulerHelper.ts
```diff
@@ -116,6 +117,37 @@ export class SlotSchedulerHelper {
...
+    for (const program of slotPrograms) {
+      if (!Number.isFinite(program.duration) || program.duration <= 0) {
+        continue;
+      }
+      for (const { customShowId } of program.parentCustomShows) {
+        schedulableShowIds.add(customShowId);
+      }
+    }
+
+    const problems = seq.collect([...zipWithIndex(slots)], ([slot, index]) => {
```
chrisbenincasa: **Save-time reference validation covers custom shows only.** The early return for any `slot.type !== 'custom-show'` means slots pointing at deleted filler lists, shows, or smart collections still pass.

Delete a filler list that a random slot references, then save that channel: 200 OK. At regeneration the slot finds no content, `warnOnUnusableFillerLists` logs it, and depending on the slot's fixed/dynamic shape the channel either plays flex for the slot's whole duration or skips it as empty. The save gave no feedback, and the strict schema cannot catch it because the reference is by id rather than by shape. Consider validating references for every slot type that carries one.

server/src/api/channelsApi.ts
```diff
@@ -584,12 +597,30 @@ export const channelsApi: RouterPluginAsyncCallback = async (fastify) => {
...
+        );
+        if (invalidDurations.length > 0) {
+          return res.status(400).send(invalidDurations.join('; '));
+        }
       }
 
-      const result = await req.serverCtx.channelDB.updateLineup(
-        req.params.id,
-        programmingRequest,
+      const updateResult = await Result.attemptAsync(() =>
```
chrisbenincasa: **The rethrow discards the original message.** `attemptAsync` wraps every failure in the anonymous message-less `WrappedError`, so `throw updateResult.error` propagates an error with an empty message for anything that is not a `BadRequestError`. On main this call propagated the original error directly.

A SQLite or drizzle failure inside the custom-show or lineup save transaction (constraint, busy database, bad enum value) now surfaces as a 500 with an empty body; the web client's error extraction yields an empty string, so the snackbar reads "Error saving schedule." with nothing after it and the operator has nothing in the response to diagnose with. Recover the original error from `.cause` before rethrowing, or rethrow `updateResult.error.cause` when it exists.

server/src/api/channelsApi.ts
```diff
@@ -584,12 +597,30 @@ export const channelsApi: RouterPluginAsyncCallback = async (fastify) => {
             slots: groupValidation.sanitizedSlots,
           },
         };
+      } else {
+        const invalidDurations = req.body.lineup.flatMap((program, index) =>
```
chrisbenincasa: **Contract contradiction.** This guard rejects any item failing `Number.isFinite(d) && d > 0` with 400 "it must be positive", but the endpoint's request schema advertises `content.duration` as `z.number().min(0)` (`types/src/api/lineupPrograms.ts`), and the regenerated OpenAPI spec still publishes `minimum: 0`.

A client generated from the published spec can send `{type: 'manual', lineup: [{type: 'content', id, duration: 0}]}` and is rejected 400 for a payload the contract explicitly allows. The web client filters zero durations before sending, so only direct API consumers meet the contradiction. Either tighten the published schema to `.positive()` (and regenerate the client) or align the guard with `min(0)`.

server/src/db/CustomShowDB.ts
```diff
@@ -145,22 +154,45 @@ export class CustomShowDB {
...
+          .set(updates)
+          .where(eq(CustomShow.uuid, show.uuid))
+          .run();
+      }
+
+      if (replacementPrograms === undefined) {
+        return;
+      }
+
+      tx.delete(CustomShowContent)
```
chrisbenincasa: **One table, two opposite conventions for `[]`.** This branch makes an explicit empty list mean "clear membership" here, but the other two writers of `custom_show_content` still treat empty input as a no-op: `CustomShowSyncService.ts:97-118` skips the upsert and only logs when the playlist returns 0 items, and `upsertCustomShowContent` returns early for an empty array (`:302-303`).

Concrete consequence: a user empties the Plex playlist a custom show syncs from and runs Sync. The log says "Got 0 items from external playlist", the show keeps its previous programs, and it keeps playing content the user removed. The count never reaches zero, the new empty-show availability warnings never fire, and the only route to the intended empty state is deleting and recreating the show. Pick one convention and apply it to all three writers.

server/src/db/CustomShowDB.ts
```diff
@@ -125,11 +126,19 @@ export class CustomShowDB {
       return null;
     }
 
-    if (updateRequest.programs && updateRequest.programs.length > 0) {
-      await this.upsertCustomShowContent(show.uuid, updateRequest.programs);
+    // An omitted field leaves membership alone. An explicit list, including an
+    // empty one, replaces it. Resolve every ID before changing anything.
+    const replacementPrograms = updateRequest.programs;
+    if (replacementPrograms !== undefined) {
```
chrisbenincasa: **The contract audit now certifies the destructive path as safe.** The KNOWN entry for `UpdateCustomShowRequestSchema.programs` in `server/src/testing/zodContractAudit.test.ts:90` still reads "defused by a length guard in CustomShowDB.saveShow". This change deletes that guard, so the audit entry is stale and the test stays green while tracking the exact data-loss path described in the comment on `saveShow` above. Update the entry (and its reasoning) to match the new replacement-branch behavior — and consider whether it should move out of KNOWN entirely.

server/src/services/scheduling/TimeSlotService.ts
```diff
@@ -104,7 +105,9 @@ export async function scheduleTimeSlots(
 
   // Load programs
   // TODO: include redirects and custom programs!
-  const allPrograms = deduplicatePrograms(programs);
+  const allPrograms = partitionSchedulablePrograms(
```
chrisbenincasa: **The time path drops the unschedulable bucket silently.** The random path keeps it and warns once per generation (`RandomSlotsService.ts:238-242`, "Skipping N program(s) without a positive duration"), so an operator debugging "my channel only shows flex" sees a warning for one slot type and silence for the other.

A channel whose content all has zero or non-finite durations schedules as pure flex with nothing in the logs, and the operator cannot distinguish a broken library from an intentionally empty slot. Log the dropped count here the same way the random path does.

web/src/components/slot_scheduler/EditSlotProgrammingForm.tsx
```diff
@@ -49,11 +52,18 @@ export const EditSlotProgrammingForm = <
...
-    return map(
-      uniqBy(programOptions, ({ type }) => type),
-      'type',
+    const types = new Set<ProgramOptionType>(
+      programOptions.filter(isSelectableForNewSlot).map(({ type }) => type),
     );
-  }, [programOptions]);
+    // A saved slot keeps its own type listed so it stays editable after its
+    // content becomes unavailable.
+    types.add(type);
```
chrisbenincasa: **A listed type with no options can throw on click.** Keeping the slot's own type in the Select is right, but `handleTypeChange` (`:71-80`) calls `newSlotForType` at `:78` before `reset` at `:79` can run, and `newSlotForType` throws for an optionless custom-show slot (`EditRandomSlotDialogContent.tsx:284-288`) and dereferences undefined for filler/show.

Link a slot to a slot whose type has no selectable content: `SlotLinkingControl` resets the form's `type` while the local `typeSelectValue` goes stale, leaving that type visible in the Select. Clicking it throws inside the Select's `onChange`, React does not catch handler errors, the form never resets, and the user sees only a console error with a stuck dialog. Not reachable on main, where a listed type always had a matching option.

server/src/db/channel/LineupRepository.ts
```diff
@@ -973,6 +974,11 @@ export class LineupRepository {
         }
       });
 
+      await this.assertProgramsExist(
```
chrisbenincasa: **`append: true` gap.** The existence check receives `req.lineup.map(...)` — the incoming items — rather than the merged `newLineupItems`, so when appending, only the appended items are validated and the channel's existing items are rewritten unchecked. The new manual-duration guard in `channelsApi.ts:600-610` has the identical gap.

A manual-lineup append to a channel whose stored lineup references a program deleted since that lineup was written succeeds and persists the dead reference, leaving an item in the lineup that can never resolve. That is the case this check was added to prevent, one write path over. Validate the merged lineup.

server/src/api/customShowsApi.ts
```diff
@@ -105,16 +107,25 @@ export const customShowsApiV2: RouterPluginAsyncCallback = async (fastify) => {
...
+          400: z.string(),
           404: z.void(),
         },
       },
     },
     async (req, res) => {
-      const customShow = await req.serverCtx.customShowDB.saveShow(
-        req.params.id,
-        req.body,
+      const saveResult = await Result.attemptAsync(() =>
```
chrisbenincasa: **Pre-existing, but adjacent to this branch's error plumbing: the DELETE 404 is dead.** The DELETE handler at `:263` tests `isNull(customShow)`, but `customShowDB.getShow` returns `undefined` for a missing id (its own test asserts this), so the 404 branch never fires and `deleteShow` runs for ids that don't exist. A client probing existence before deleting gets a success status for a show that was never there, and any non-idempotent work later added to that branch would run against a nonexistent target. Since this branch is adding the 400/`findBadRequestError` handling here anyway, switch the check to `isNil` while you're in the file.

chrisbenincasa (commented):
15 inline findings, ranked by severity. The top four were reproduced by execution on this branch. (Posted as COMMENT — GitHub won't let the PR author request changes on their own PR.)

**Blocking**

1. **Data loss on custom-show update.** `UpdateCustomShowRequestSchema` is `.partial()` of a schema whose `programs` has a default of `[]`, so a PUT that omits `programs` takes the new replacement branch and deletes every `custom_show_content` row. The web form hits this on every save of a synced show.
2. **The new 400s never fire in production.** `Result.attemptAsync` wraps every error in an anonymous `WrappedError`, so the `instanceof TypedError` checks in `TunarrWorker` are always false and the worker pool (on by default) rejects with an empty 500. Tests stay green because vitest forces `TUNARR_USE_WORKER_POOL=false`.
3. **The #2087 hang is still persistable via time slots.** The time save path runs only `validateSlotGroups`; a fractional `startTime` saves with 200 and then hangs the scheduler in the same non-advancing loop class this PR fixes for random slots. Reproduced: loop blocked the event loop, killed at 60s.
4. **Regeneration regression.** `validateSchedule()` runs unconditionally, so schedules already in the DB with `durationMs: 0` or fractional `programCount` fail every regeneration with an uncontextualized 500. On main they played with flex padding.

**Should fix in the same pass**

5. The editor cannot repair a flex/redirect slot carrying a dynamic durationSpec — the control that would fix it is hidden, so the new 400 is a dead end.
9. The manual-lineup guard rejects `duration: 0` while the published spec says `minimum: 0` — contract contradiction for API consumers.
11. The contract-audit KNOWN entry still cites the deleted `programs.length > 0` guard, so the audit now certifies the destructive path as safe.
13. A stale type option in the slot editor throws inside React's onChange after slot linking resets the form.

**Also noted**: web/server disagree on "empty custom show" (6); reference validation covers custom shows only (7); `attemptAsync` + rethrow swallows original error messages (8); `[]` means "clear" in `saveShow` but "no-op" in the other two writers of the same table (10); time path drops unschedulable programs silently while the random path warns (12); `append: true` validates only appended items (14); dead 404 in the DELETE handler (15, pre-existing).

PR #2112, opened 2026-09-17
fix(server): relabel Jellyfin/Emby grouping external IDs stored as plex

### LINKED ISSUE #2084: Guide, lineup, and programming endpoints return 500 when a Jellyfin/Emby grouping's external ID is labeled plex
## Summary

Some Jellyfin and Emby shows and seasons have their only external-ID row stored with `source_type = 'plex'`. `ApiProgramConverters.convertProgramGrouping` can't find an external ID that matches the grouping's own source, so it throws. Any channel whose lineup contains one of those episodes then returns 500 from several endpoints.

## Symptoms

These endpoints return 500 for an affected channel:
- `GET /api/guide/channels`, which fails for every channel if any one channel is affected
- `GET /api/channels/:id/programming`
- `GET /api/channels/:id/lineup`

```
{"statusCode":500,"error":"Internal Server Error","message":"Program grouping <uuid> has no external ID!"}
```

The error is thrown at `server/src/api/ApiProgramConverters.ts:276`. It reproduces on `dev` (a9501647) and on `feat/zod-4.5-compile`, so it isn't related to the Zod upgrade.

## Evidence from one real DB

Mislabeled rows in `program_grouping_external_id`:

| `source_type` | `external_source_id` | Grouping source | Type | Rows | Created |
|---|---|---|---|---|---|
| `plex` | `Jellyfin` | `jellyfin` | season | 8 | 2025-09-16 18:41 |
| `plex` | `Jellyfin` | `jellyfin` | show | 5 | 2025-09-16 18:41 |
| `plex` | `Emby` | `emby` | season | 4 | 2025-09-19 20:20 |
| `plex` | `Emby` | `emby` | show | 3 | 2025-09-19 20:20 |

- Every Emby grouping in this DB is affected (7 of 7). Only 13 of 699 Jellyfin groupings are.
- The bad rows have the correct `media_source_id`. `BackfillMediaSourceIdFixer` matches on `sourceType` and couldn't have set it from a `plex` label. So whatever inserted these rows wrote `plex` directly.
- The rows were created in two bursts, three days apart, in September 2025. I haven't checked whether current scanner or minting code still writes this.

## Detect

```sql
SELECT g.source_type, g.type, count(*)
FROM program_grouping g
WHERE g.source_type IN ('jellyfin', 'emby')
  AND NOT EXISTS (
    SELECT 1 FROM program_grouping_external_id e
    WHERE e.group_uuid = g.uuid AND e.source_type = g.source_type
  )
GROUP BY 1, 2;
```

## Possible fixes

- Find the write path that set `source_type = 'plex'` for Jellyfin/Emby grouping external IDs, and confirm whether it's still live.
- Add a fixer that relabels rows where `source_type = 'plex'` and `external_source_id` names a Jellyfin or Emby media source.
- Decide whether the converter should skip or degrade one bad grouping instead of failing the whole guide response.

Lingikaushikreddy:
Fixes #2084.

## What

Adds `FixMislabeledGroupingExternalIdSourceType`, a startup fixer registered in `FixerModule`. It relabels `program_grouping_external_id` rows that are stored as `source_type = 'plex'` but whose `media_source_id` points at a Jellyfin or Emby media source, using that media source's type.

With the rows relabelled, `ApiProgramConverters.convertProgramGrouping` finds the external ID for the grouping's own source again, so `/api/guide/channels`, `/api/channels/:id/lineup` and `/api/channels/:id/programming` stop returning 500 for affected channels.

## Guard rails

- It only touches rows with a non-null `media_source_id` whose media source is `jellyfin` or `emby`. Rows from real Plex sources are left alone.
- If the grouping already has a correctly labelled `jellyfin`/`emby` row for the same media source, the mislabelled row is skipped. Relabelling it would violate `unique_program_grouping_multiple_external_id_media_source`, and the correct row already satisfies the converter.
- It's a single `UPDATE ... WHERE` in the same style as `BackfillMediaSourceIdFixer`, so it's a no-op on databases without the bad rows.

Not in this PR: finding the September 2025 write path that produced the rows (the report notes it may no longer be live), and making the converter degrade instead of throwing. Happy to follow up on either if you'd like.

## Tests

`FixMislabeledGroupingExternalIdSourceType.test.ts` runs against a pre-migrated database, like `MediaSourceDB.test.ts`:
- relabels plex-labelled rows for a Jellyfin source and an Emby source
- leaves rows from a real Plex source alone
- skips a mislabelled row when a correct row exists for the same grouping and media source (no unique-index error)

Locally on `main`: `pnpm typecheck` (server) is clean, `oxlint` is clean, Prettier applied. The full server suite passes except `ArtworkService.test.ts` › "never redirects a credentialed URL, even with proxyArtwork disabled", which fails the same way on unmodified `dev` for me.


Lingikaushikreddy:
Small correction on my testing note. I had checked the unrelated `ArtworkService.test.ts` failure once. I've since run it twice on unmodified `dev` at 39bc7c2, without this branch: `1 failed | 5 passed` both times, the same "never redirects a credentialed URL, even with proxyArtwork disabled" case. So it is reproducible on `dev` and not caused by this PR.

chrisbenincasa (changes requested):
## Review

Verified by checking out the branch and running it locally: the new tests pass (3/3), the full `src/tasks/fixers/` folder passes, `oxlint` is clean on the changed files, and a full `tsc -p tsconfig.build.json --noEmit` shows no errors in any PR file.

The approach is correct: the UPDATE is properly scoped (guards key off the row's `media_source_id` pointing at a jellyfin/emby source, so legitimate cross-source Plex rows are untouched), the `NOT EXISTS` guard matches the `(group_uuid, source_type, media_source_id)` unique index exactly, and startup ordering is safe — `GenerateGuideStartupTask` depends on `FixerRunner`, so the guide is generated after the relabel. This aligns with the root cause of #2084: the mislabeling write path was removed in `c95d8028a` (2026-04-10), so no new bad rows are produced and data remediation is the right shape for the fix.

One substantive inline comment (NULL `media_source_id` rows from the Jan–Feb 2025 window are missed), plus two nits. Worth tracking separately: hardening `convertProgramGrouping` so a future bad grouping degrades instead of 500ing the whole `/api/guide/channels` response.

server/src/tasks/fixers/FixMislabeledGroupingExternalIdSourceType.ts
```diff
@@ -0,0 +1,90 @@
...
+              '=',
+              'programGroupingExternalId.mediaSourceId',
+            )
+            .where('mediaSource.type', 'in', ['jellyfin', 'emby'])
+            .select('mediaSource.type')
+            .$narrowType<{ type: 'jellyfin' | 'emby' }>()
+            .limit(1),
+      })
+      .where('programGroupingExternalId.sourceType', '=', 'plex')
+      .where('programGroupingExternalId.mediaSourceId', 'is not', null)
```
chrisbenincasa, line 49: Requesting a change here: this `mediaSourceId IS NOT NULL` filter leaves behind the oldest variant of these bad rows. The buggy mint path was live from `3a96d3d91` (2025-01-21, #910), but `media_source_id` wasn't written on grouping external IDs until `769b05d20` (2025-02-28, #1106). Rows minted in that window have `source_type = 'plex'`, `external_source_id` = the media source name ('Jellyfin'/'Emby'), and `media_source_id = NULL` — and `BackfillMediaSourceIdFixer` can never backfill them because it matches `mediaSource.type = sourceType`, which fails for `plex` + 'Jellyfin'. (Emby has a narrower but real window too: Emby support landed 2025-02-18.)

Such rows break the converter exactly the same way, so databases from that window would still 500 after this fixer runs. Please either add a second UPDATE clause for `media_source_id IS NULL` rows whose `external_source_id` matches a jellyfin/emby media source name (with the same NOT EXISTS guard), or, if you'd rather keep this PR scoped, file a follow-up issue and extend the issue's detect SQL to cover the NULL-`media_source_id` case.

server/src/tasks/fixers/FixMislabeledGroupingExternalIdSourceType.ts
```diff
@@ -0,0 +1,90 @@
...
+                'existing.mediaSourceId',
+                '=',
+                'programGroupingExternalId.mediaSourceId',
+              )
+              .where('existing.sourceType', 'in', ['jellyfin', 'emby'])
+              .select('existing.uuid'),
+          ),
+        ),
+      )
+      .executeTakeFirst();
```
chrisbenincasa: Nit: `.execute()` (like `BackfillMediaSourceIdFixer`) would be more idiomatic than `executeTakeFirst()` for an un-LIMITed UPDATE — both return `UpdateResult` with `numUpdatedRows`.

server/src/tasks/fixers/FixMislabeledGroupingExternalIdSourceType.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      'plex',
+    );
+    const embyRow = await insertExternalId(
+      db,
+      await insertShow(db, emby),
+      emby,
+      'plex',
+    );
+
+    await new FixMislabeledGroupingExternalIdSourceType(db).run();
```
chrisbenincasa, line 122: Nit: fixers run on every startup and must be idempotent (`FixerRunner` comment). Worth locking that in with a second `run()` + assertions in this test — cheap, and it covers part of the fixer contract.

server/src/tasks/fixers/FixMislabeledGroupingExternalIdSourceType.ts
Lingikaushikreddy: Fixed in 2f4b90e. Added a second UPDATE for `media_source_id IS NULL` rows, matching `external_source_id` against a jellyfin/emby media source name, with the same NOT EXISTS guard (keyed on the null-media-source rows for that grouping, matching the `unique_program_grouping_single_external_id_media_source` index).

Thanks for the archaeology on the window — I had only looked at rows the first statement could see.

Two notes on the shape:
- It relabels only; `media_source_id` stays null. `BackfillMediaSourceIdFixer` binds before this one in `FixerModule`, and once the label is right its `mediaSource.type = sourceType` join matches, so the next startup fills the id in. Happy to set it here instead if you would rather it be one pass.
- The guard uses `sourceType in ('jellyfin','emby')` rather than resolving the exact target type, same as the first statement. Conservative: at worst it skips a row that would not actually have collided.

Tests: two for the null-media-source path (relabelled, and skipped when the grouping already has a correct row) plus one that a plex-source name is left alone. They fail on the previous commit and pass now.

server/src/tasks/fixers/FixMislabeledGroupingExternalIdSourceType.ts
Lingikaushikreddy: Done in 2f4b90e — both statements use `execute()` now.

server/src/tasks/fixers/FixMislabeledGroupingExternalIdSourceType.test.ts
Lingikaushikreddy: Added in 2f4b90e: "is idempotent: a second run changes nothing" runs the fixer twice over one row of each kind and asserts both stay relabelled. It fails on the previous commit, since the null-media-source row was never relabelled in the first place.

Lingikaushikreddy:
Pushed 2f4b90e with all three: the null-`media_source_id` window, `execute()`, and an idempotency test. Replies inline.

Verification on this branch: `vitest src/tasks/fixers/` 7 passed (the three new cases fail on the previous commit), `oxlint` clean, `tsc -p tsconfig.build.json --noEmit` 0 errors, Prettier applied.

On hardening `convertProgramGrouping` so one bad grouping degrades instead of failing the whole `/api/guide/channels` response: happy to open an issue for that, or take it in a follow-up PR if you would rather have it tracked with a fix attached. Your call on which.

chrisbenincasa (commented):
## Re-review (`2f4b90e`)

All three items from the previous review are addressed: the second `UPDATE` for `media_source_id IS NULL` rows (`FixMislabeledGroupingExternalIdSourceType.ts:90-137`), the `execute()` switch, and the idempotency test. Verified locally on `2f4b90ed`:

- new test file: **7/7 pass**
- `tsgo -p tsconfig.build.json --noEmit`: **clean**
- `oxlint` on both new files: **0 warnings / 0 errors**
- Confirmed `ApiProgramConverters.convertProgramGrouping` (`ApiProgramConverters.ts:270-277`) keys on `eid.sourceType === mediaSource.type`, so relabeling alone resolves the throw and the one-startup-deferred `media_source_id` backfill does not affect the reported endpoints. The `FixerModule` ordering note is accurate.

### Substantive: the name match can relabel a legitimate Plex row

The new clause (`:90-137`) matches `external_source_id` against the name of any jellyfin/emby source, with no check that a source of the row's **current** type shares that name, and no correlation to the grouping's own source.

`media_source` is only unique on `(type, name, uri)` (`Migration1741658292_MediaSourceIndex.ts:6-12`), so a Plex source and a Jellyfin/Emby source **may share a name**. Reproduced against the pre-migrated test DB:

- plex source `Shared Name`, jellyfin source `Shared Name`
- show grouping owned by the **plex** source, external-id row `source_type='plex'`, `media_source_id=NULL`, `external_source_id='Shared Name'`
- after the fixer: `source_type='jellyfin'` — relabeled away from its real type

That inverts the failure mode: a Plex grouping loses its plex-typed ID, so `convertProgramGrouping` for the Plex source can then hit the same "has no external ID!" throw this PR fixes, and `ProgramGroupingRepository.getProgramGroupingByExternalId` (uuid-keyed) misses it.

Cheap guard — skip a candidate when a source of its own type also owns that name, or require the grouping's own media source to match:

```ts
.where((eb) =>
  eb.not(
    eb.exists(
      eb.selectFrom('mediaSource as other')
        .whereRef('other.name', '=', 'programGroupingExternalId.externalSourceId')
        .whereRef('other.type', '=', 'programGroupingExternalId.sourceType')
        .select('other.uuid'),
    ),
  ),
)
```

Also, the `SET` subquery's `.limit(1)` (`:95-104`) picks arbitrarily if a Jellyfin and an Emby source share a name; a unique-candidate check avoids relabeling to the wrong type.

Severity is narrow: this only bites the older NULL-`media_source_id` window, and #2084's observed rows all had the correct `media_source_id`, so this clause is defensive. If you'd rather keep scope, documenting the collision assumption and tracking it is a reasonable alternative — but the guard is ~5 lines.

### Minor

1. **The NULL-path guard's rationale doesn't hold.** The reply cites `unique_program_grouping_single_external_id_media_source`. That index is `(group_uuid, source_type, media_source_id) WHERE media_source_id IS NULL` (`ProgramGroupingExternalId.ts:56-58`); because the key column is NULL for every indexed row, SQLite treats entries as distinct and the index does **not** actually enforce uniqueness (confirmed directly). So the guard prevents duplicates/ambiguity, not a constraint violation. It also only skips when the existing correct row *also* has `media_source_id IS NULL` — a correct row with a non-null id is invisible to it. Related: two relabeled `(group, jellyfin, NULL)` rows can make the next startup's `BackfillMediaSourceIdFixer` hit `unique_program_grouping_multiple_external_id_media_source` when it fills the id. Worth a corrected comment and possibly matching the first clause's keying.
2. **`mediaSourceId` deferred to next startup.** Not blocking (the converter keys on `sourceType`), but setting it in the same statement completes the repair in one pass, removes the `FixerModule` ordering dependency, and fixes uuid-keyed lookups immediately. The author offered this; my lean is to take it.

### Still open

Hardening `convertProgramGrouping` so one bad grouping degrades instead of 500-ing the whole `/api/guide/channels` response remains the durable fix (last bullet in #2084's "Possible fixes"). Out of scope here, but worth a tracking issue before #2084 is closed. Also, GitHub reports no CI checks on this branch, so the PR currently has no automated gate.

Lingikaushikreddy:
Pushed 4fbe7622 (rebased onto `main` at 921ef7f1 first). All three points addressed.

**The name match relabelling a legitimate Plex row.** You are right, and the reproduction is the useful part of the review: a name is not an identity, because `media_source` is unique on `(type, name, uri)`. Rather than only excluding a same-type namesake, the clause now requires that **exactly one** media source carries the name:

```ts
.where((eb) => eb.not(eb.exists(
  eb.selectFrom(['mediaSource as candidate', 'mediaSource as other'])
    .whereRef('candidate.name', '=', 'programGroupingExternalId.externalSourceId')
    .whereRef('other.name', '=', 'candidate.name')
    .whereRef('other.uuid', '!=', 'candidate.uuid')
    .select('candidate.uuid'))))
```

That covers your case and the one you flagged next to it, where a Jellyfin and an Emby source share a name and `.limit(1)` would pick arbitrarily. Anything ambiguous is left alone for a human rather than relabelled on a guess.

**`media_source_id` in the same statement.** Taken, as you leaned. The `SET` now writes both fields, so the repair completes in one pass and the `FixerModule` ordering is no longer load-bearing; uuid-keyed lookups work immediately rather than after the next startup.

**The guard's rationale.** My comment was wrong and your reading is right: the partial index has NULL in its key column for every indexed row, so SQLite does not enforce uniqueness there. The comment now says the guard is about ambiguity rather than a constraint, and the guard itself is widened to any `jellyfin`/`emby` external id on the grouping regardless of its `media_source_id`, so a correct row with a non-null id is no longer invisible to it. That also removes the path you described where two relabelled `(group, jellyfin, NULL)` rows would collide once `BackfillMediaSourceIdFixer` filled the ids.

**Tests:** four new cases, all of which fail on the previous commit and pass now — a Plex source sharing the name, a Jellyfin and an Emby sharing it, `media_source_id` set in the same pass, and a correct row with a non-null id blocking the relabel. 11 pass in the file.

Locally: `tsgo -p tsconfig.build.json --noEmit` clean, `oxlint` clean on both files (the `AddOverflowMigration.ts` finding also appears on unmodified `main`), Prettier applied.

**Still open, as you say:** hardening `convertProgramGrouping` to degrade instead of 500-ing. Happy to open that tracking issue and take it after this lands, if you want it separate from #2084.

On CI: the checks show as not run — I think they need a maintainer approval on this branch.

PR #2121, opened 2026-09-20
feat: Allow users to use ersatztv-next as their streaming pipeline

chrisbenincasa:


chrisbenincasa:
One note from the commit-by-commit review, about the 0048 migration and an existing mechanism that fits it.

The drizzle rebuild pattern (`PRAGMA foreign_keys=OFF` ... `DROP TABLE channel` ...) is dangerous in this repo's migration path. `DrizzleMigrator` runs sql-file migrations via `upDrizzle`, which `wrapWithTransaction` wraps in a native better-sqlite3 transaction. `PRAGMA foreign_keys` is a no-op inside a transaction, and `DBAccess` enables FK enforcement per connection. Five tables FK into `channel` with ON DELETE CASCADE (channel_programs, channel_fallback, channel_filler_show, program_play_history, channel_subtitle_preferences), so `DROP TABLE channel` under FK enforcement fires an implicit DELETE that cascades into all five. The channel rows survive the `__new` copy, but every lineup row is gone. I verified this empirically with better-sqlite3: same statement sequence inside a transaction, the child table ended at 0 rows. (The prior channel rebuild, 0040, shipped before the synchronous-transaction migrator landed, which is why it never hit this.)

The fix for exactly this shape already exists: `fullCopy`. `makeMigrationFromSqlFile('./sql/0048_...sql', true)` routes the migration through `DatabaseCopyMigrator`, which runs the migration against a temp copy of the DB and then re-copies all data from the intact original before swapping the file over the live path. The cascade wipe becomes harmless because it happens in the throwaway copy. That's how `Migration1738604866_AddEmby` and `Migration1756381281_AddLibraries` handle their rebuilds.

0048 is deleted later in this branch, so there's nothing to change now, but two things worth carrying forward:

- A channel-table rebuild (or any rebuild of a table with inbound FKs) needs `fullCopy: true`, or the migration infrastructure needs autocommit/`noTransaction` execution for PRAGMA-carrying migrations, which would be the durable fix.
- The underlying trap applies to every future sql-file migration, not just this one.

chrisbenincasa (commented):
## Review: ersatztv-next streaming pipeline (all 29 commits, base 7c4d08d7 → head 9ce5854ee)

**Verdict: approve.** Remaining findings are all minor; none block.

### Findings (10 total)

Resolved in-PR:
- **F1 [critical] — resolved at commit 14.** The `channel` rebuild migration would have silently wiped five lineup tables (or bricked startup) under the transaction-wrapped sql-file migrator with FK cascades on. Migration deleted; the stream mode rides on the feature flag instead.
- **F3 [major] — verified at commit 16.** `sample_rate_hz` was emitted in kHz, 1000x off; worker died on real streams. Fixed to `audioSampleRate * 1000`, independently confirmed by the commit's own failure trace.
- **F10 [minor] — fixed at commit 29.** Stock VAAPI channels silently lost all hardware accel under etv_next: the worker requires both `vaapi_device` and `vaapi_driver`, Tunarr's defaults supply neither, and `system` was exempted from warnings. Device now defaults to `/dev/dri/renderD128` and `system` is inferred from the render node's PCI vendor id; where neither can be determined the compatibility notice says the channel will transcode in software. Upstream filed as ErsatzTV/next#246.

Open minors:
- **F2**: the schema-render "closed subset" guarantee only covers `type`; other unsupported JSON Schema keywords are silently dropped, not thrown.
- **F4**: several per-config settings are dropped without being named, against the mapper's own "dropped and named" contract.
- **F5**: subtitles are silently dropped in the etv path either way.
- **F6**: a partial materialized-window rebuild rewrites the window with a shrunk `finishMs`, discarding the healthy tail until the next tick.
- **F7**: the resolve path never passes the per-session token the integration doc requires, so `StreamProgramCalculator`'s too-many-attempts throttler never fires on the etv path — a failing program retries once per 44s chunk for its whole duration instead of degrading to Tunarr's throttled error screen. Safe to fix: `TOO_FREQUENT` is 100ms vs the worker's 44s cadence, so a token cannot misfire.
- **F8**: Emby stream tokens survive troubleshoot redaction — the pattern `X-Emby-Token:\s*` matches only the Jellyfin header form, not Emby's `?X-Emby-Token=` URI form, which leaks through the worker's `optimized pipeline` line, the stderr tail, and the dossier report. Plex and Jellyfin are covered. Same patterns pre-exist in `TroubleshootService.ts`. Fix: accept both separators, e.g. `X-Emby-Token[:=]\s*`.
- **F9**: the docs' "Programming edits land within half an hour" paragraph describes the materialized window mode, which no production path uses. The default dynamic mode resolves every item against Tunarr, so an edit lands at the next item (one ≤44s chunk), and can restart the currently-airing program — the opposite of "leaves the program currently airing alone". The bound itself holds; the mechanism text misdescribes the default.

### Tests run

Per-commit vitest in a clean worktree: full suite 250→258 passing through the middle commits; SessionManager 14/14; mapper 59/59; troubleshoot suite included in 236/236; remux fix 28/28; VAAPI fix 40/40. Full server suite at head: **1852 passed / 2 skipped** (144 files), exit 0. oxlint clean on touched files; `turbo typecheck` 5/5 at commit 13.

### Recommended before the flag stops being experimental

F7 (throttler parity), F8 (token in shared troubleshoot dossiers), and F10's documented residual (a wrong driver inference falls back to software silently — no regression, but worth the upstream fix landing). F2/F4/F5/F6/F9 are doc/polish class.

Issue #2132, opened 2026-09-21
Write transactions use DEFERRED, so busy_timeout never applies to the lock upgrade


chrisbenincasa:
> *This was generated by AI during triage.*

**Category:** code health / hardening

## Summary

Every write transaction in the server uses SQLite's default `DEFERRED` behavior. A deferred transaction that upgrades from reader to writer mid-way gets `SQLITE_BUSY` **immediately, without the busy handler ever waiting** — so the 5-second `busy_timeout` the connection is configured with does not protect the case most likely to hit it. Switching write paths to `BEGIN IMMEDIATE` closes that hole.

## Current state

`DBAccess` opens the connection with a 5-second busy timeout (`server/src/db/DBAccess.ts:46-48`):

```ts
this.sqlite = new Sqlite(name, {
  timeout: 5000,
});
```

Drizzle's better-sqlite3 session picks the transaction variant from a `behavior` option and defaults to deferred:

```js
return nativeTx[config.behavior ?? "deferred"](tx);
// drizzle-orm/better-sqlite3/session.js:34
```

There are **32 `.transaction(` call sites across 13 files** in `server/src` (excluding tests), and **none of them passes a `behavior`**. `grep -rn "behavior" server/src` returns no transaction-related match. So every write runs deferred.

better-sqlite3 11.8.1 does expose the variants — `deferred`, `immediate` and `exclusive` are all defined at `better-sqlite3/lib/methods/transaction.js:15-17,43-45` — so this is purely a matter of passing the option.

## Why deferred is the wrong default here

A `DEFERRED` transaction acquires no lock at `BEGIN`. It starts as a reader and only tries to take the write lock at its first write statement. If another connection has written in the meantime, SQLite returns `SQLITE_BUSY` at that upgrade point and **will not invoke the busy handler**, because blocking there risks deadlock. The timeout does not apply.

Two consequences:

1. The 5000 ms `busy_timeout` protects the case where a transaction takes the write lock at `BEGIN`, but not the deferred upgrade — which is the path all 32 sites currently take.
2. The failure happens *after* the transaction has already read and possibly done work, so the caller has more to unwind.

`BEGIN IMMEDIATE` takes the write lock up front, where the busy handler is allowed to wait. Contention becomes a bounded wait instead of an immediate throw.

## Current exposure

Low today, because there is normally only one writer:

[...]

PR #2133, opened 2026-09-21
fix(artwork): derive artwork urls on demand and bound the backfill

### LINKED ISSUE #2128: BackfillProgramArtworkFixer rescans the whole library on every boot and blocks the main thread
> *This was generated by AI during triage.*

## Summary

`BackfillProgramArtworkFixer`, added in #1923, rescans the entire `program` and `program_grouping` tables on **every server start**, and does all of that work synchronously on the main thread. It has been in stable releases since v1.3.11 (2026-08-10).

## Why `canRunInBackground` doesn't help

The class sets `canRunInBackground = true` (`server/src/tasks/fixers/BackfillProgramArtworkFixer.ts:19`), which makes `FixerRunner` skip the `await` and let the fixer settle on its own:

```ts
if (!fixer.canRunInBackground) {
  await fixerPromise;          // FixerRunner.ts:38-39
}
```

That flag can't do anything here. `runInternal` (`:27`) calls `backfillProgramArtwork()` and `backfillGroupingArtwork()` straight through and only then returns an already-resolved promise (`:41`):

```ts
protected runInternal(): Promise<void> {
  const programCount = this.backfillProgramArtwork();
  const groupingCount = this.backfillGroupingArtwork();
  ...
  return Promise.resolve(void 0);
}
```

Both helpers are synchronous drizzle/better-sqlite3 calls, so all the work completes *before* the promise exists. The runner's background branch gets a settled promise and never yields. This is not fixable by reshuffling the flag — better-sqlite3 is synchronous by design, so no sqlite-backed fixer can meaningfully run "in the background" on the main thread.

## Why it re-runs forever

`FixerRunner.runFixers` (`server/src/tasks/fixers/FixerRunner.ts:28-29`):

```
// Fixers currently do not keep any state and we will
// just run them at each server start. As such, they
// should be idempotent.
```

There is no version gate and no completion marker. Registered unconditionally at `FixerModule.ts:12`. So every boot pays two full `LEFT JOIN` scans of `program x artwork` and `program_grouping x artwork` even when there is nothing left to backfill. The inserts are chunked at 100 (`:132`, `:248`) but that is only JS array chunking inside one `drizzleDB.transaction` (`:131`, `:247`) — it does not bound the scan or release the loop.

## Impact

On a large library (100k+ programs) the first boot does two full scans plus ~100k inserts with no yield point. Startup, HTTP handling and any in-flight stream stall for the duration. Later boots skip the inserts but still pay both scans, every time, permanently.

## Suggested fix

Bound the work per run rather than trying to make it async:

1. **Make it incremental.** Process a fixed budget per boot (say N rows) and persist a resume cursor, instead of scanning the whole table each time.
2. **Add completion state.** Fixers need somewhere to record "done" so a finished backfill stops scanning. This is a `FixerRunner` gap, not just this fixer's — the comment at `FixerRunner.ts:28` explicitly acknowledges there is no state today.
3. **Yield between batches** so a long backfill cannot monopolise the loop, or move it off the main thread entirely.
4. Consider dropping `canRunInBackground` from this fixer, since it is misleading — it promises non-blocking behaviour the sync driver cannot deliver.

## Secondary

`buildArtworkSourcePath` (`:106`, `:222`) synthesises a `sourcePath` for any program with the required fields. A program whose source genuinely has no artwork gets a persisted row pointing at a URL that 404s, and nothing later clears it.

## Environment

- Introduced by #1923, which landed on `main` as `b85dddf1` and on `dev` as `1224b2f37` — the same change as two separate commits.
- **Shipped in stable since v1.3.11 (2026-08-10)** and in every release since, including v1.3.15. This runs on every stable user's server today; it is not a pre-release issue and does not gate #1891.
- The fixer file is byte-identical on `main` and `dev`, so a fix on `main` merges cleanly into `dev`.

### LINKED ISSUE #2129: Artwork consolidation (#1923) drops XMLTV thumbnails and 404s /api/programs/:id/thumb
> *This was generated by AI during triage.*

## Summary

The artwork consolidation in #1923 (commit `1224b2f37`) removed the `/thumb` endpoint and the XMLTV fallback that depended on it. Programs without a persisted `artwork` row now emit **no** `<icon>` and **no** `<image>` in the guide at all, and previously-working thumbnail URLs 404. It has been in stable releases since v1.3.11 (2026-08-10).

## 1. XMLTV silently drops artwork

`XmlTvWriter.resolveArtworkUrl` used to always return a URL, falling back to the `/thumb` route when there was no stored artwork. #1923 replaced that whole fallback block with `return undefined`:

```diff
-  ): string {
+  ): string | undefined {
-    // Fallback to /thumb for programs with no stored artwork
-    ...
-    return `{{host}}/api/programs/${idToUse}/thumb${queryPart}`;
+    return undefined;
```

and made the caller conditional (`server/src/services/XmlTvWriter.ts:329-331`):

```ts
if (url) {
  partial.image = [{ _value: url, size: 3, type: 'poster' }];
  partial.icon = [{ src: url }];
}
```

So any program with no `artwork` row produces a guide entry with no artwork elements. Previously every program got a working URL. This is why `BackfillProgramArtworkFixer` exists (see #2128) — but the backfill only covers programs with a remote media source and the fields needed to synthesise a path, so anything it misses stays permanently artwork-less in the guide.

The `track` handling is also lost. The old fallback redirected to the album's uuid for tracks (`idToUse = program.album.uuid`); nothing replaces that.

## 2. Removed routes 404 with no alias

| Removed | Was at | Replacement |
|---|---|---|
| `GET /api/programs/:id/thumb` | `programmingApi.ts:700` | `GET /api/programs/:id/artwork/:artworkType` |
| `GET /api/metadata/external` | `metadataApi.ts:81` (file deleted) | `GET /api/programs/:id/artwork/:artworkType`, `GET /api/credits/:id/artwork/:artworkType` |

There is no redirect or compatibility alias. On upgrade these return 404.

`/thumb` in particular has been embedded in published output: every XMLTV file Tunarr has ever written contains `{{host}}/api/programs/<uuid>/thumb` URLs. A PVR that cached or re-serves that guide keeps requesting a dead path. Anyone scripting against the API, or with a bookmarked thumbnail URL, breaks the same way.

## Reproduction

1. Run a build with #1923 against a database with programs that have no `artwork` rows.
2. Fetch `/api/xmltv.xml`.
3. Affected `<programme>` entries have no `<icon>` and no `<image>`.
4. Request `/api/programs/<any-uuid>/thumb` -> 404.

## Expected behavior

Programs keep a usable guide thumbnail across the upgrade, and URLs published in already-distributed XMLTV keep resolving.

## Suggested fix

Some combination of:

1. **Restore a fallback in `resolveArtworkUrl`** so it never returns `undefined` when a URL can be derived, rather than relying on the backfill having run. Include the `track` -> album uuid case.
2. **Alias `/api/programs/:id/thumb`** to the new artwork route (302 or internal rewrite) so published guides and bookmarks keep working. Deprecate it in docs rather than deleting it outright.
3. If the old paths really are going away, call it out as a breaking change in the release notes for whichever stable release carries #1891.

## Environment

- Introduced by #1923, which landed on `main` as `b85dddf1` and on `dev` as `1224b2f37` — the same change as two separate commits.
- **Shipped in stable since v1.3.11 (2026-08-10)** and in every release since, including v1.3.15. This is a live regression, not a pre-release issue, and it does not gate #1891.
- `XmlTvWriter.resolveArtworkUrl` returns `string | undefined` on `main` as well (`XmlTvWriter.ts:258`, guard at `:246`), and `/programs/:id/thumb` is absent from `programmingApi.ts` on `main`.
- Related: #2128 (the backfill fixer this regression depends on)

## Why no one has reported it

A search of issues since 2026-08-01 turns up no reports of missing thumbnails or guide images. The most likely reason is that `BackfillProgramArtworkFixer` runs on every boot and populates the missing `artwork` rows, so most installs never see the gap — it is masking this regression at the cost of the every-boot full-table scan in #2128.

**That couples the two issues.** Making the fixer incremental or run-once (#2128) without restoring a fallback here could expose missing artwork on any install whose backfill has not finished. The fallback in `resolveArtworkUrl` should be restored before, or in the same change as, the fixer work.

chrisbenincasa:
Fixes #2129. Fixes #2128.

## Why

#1923 routed artwork through stored `artwork` rows and removed the `/thumb` fallback. `XmlTvWriter.resolveArtworkUrl` went from always returning a URL to returning `undefined`, and the caller now skips `<icon>` and `<image>` entirely when it does. An item with no artwork row loses its guide thumbnail.

That made `BackfillProgramArtworkFixer` load-bearing, and the backfill is expensive. It reads `program` and `program_grouping` in full with `.all()` on **every** server start — `FixerRunner` has no completion state — and wraps every insert in one transaction. better-sqlite3 is synchronous and nothing yields, so the scan blocks the event loop and stalls in-flight streams. `canRunInBackground = true` has no effect, because `runInternal` finishes all its work before the promise it returns exists.

Both have been in stable since **v1.3.11 (2026-08-10)**. They are live, not pre-release. Nobody has reported missing thumbnails, most likely because the every-boot backfill has been masking the regression — which is exactly why the two need fixing together.

## What changed

**The XMLTV writer derives when nothing is stored.** `resolveArtworkUrl` builds a candidate list (show poster, episode, album, the program itself) and returns the first with a matching stored row. It now walks that same list a second time when none of them has one, and emits `{{host}}/api/programs/:id/artwork/poster` for the first candidate whose source is derivable. Reusing the candidate list keeps the precedence identical one layer later, which also restores the `track` → album-uuid case the old `/thumb` fallback covered.

The guard mirrors `ArtworkService.deriveArtworkFromSource` — a source type `buildArtworkSourcePath` handles, an external key, a media source id. When those are missing the endpoint would 404, so the `<icon>` is dropped as before rather than pointing at a dead URL. No DB lookup happens at emit time and the writer stays synchronous; the endpoint does the derivation on request.

**`ArtworkService` derives on demand.** When an item has no artwork row, `resolveProgramArtwork` falls through to `deriveArtworkFromSource`, which builds the remote URL from the item's own media source and hands it to the existing `artworkToResult`. This is what the URL above resolves against, and it also covers direct `/api/programs/:id/artwork/:type` requests. The access token still travels as a request header rather than on the URL, unchanged.

Together these make the backfill a way to *persist and cache* artwork paths rather than a prerequisite for having any, so a partial backfill is not user-visible.

**The backfill is bounded by rows read.** With that safety net in place:

- Both scans page by `uuid` cursor in batches of 500 instead of `.all()` over the table.
- Each chunk of 100 inserts gets its own transaction, rather than one transaction around the whole loop.
- The loop yields between batches via the existing `util/throttle.ts`.
- A cap of 5k rows **read** per run spreads a large backfill over several restarts. Counting reads is what makes it a bound — a write counter sits at zero for a batch that derives nothing, which would let a table of unresolvable rows page to the end while inserting nothing.
- The two scans share one allowance, so a boot reads 5k rows total rather than 5k each.

No cursor is persisted. Rows drop out of the `isNull(Artwork.uuid)` predicate once written, so a restart resumes on its own.

**Write transactions use `BEGIN IMMEDIATE`.** Drizzle defaults to `DEFERRED`, where a reader-to-writer lock upgrade returns `SQLITE_BUSY` without the busy handler being allowed to wait — so the connection's 5s `busy_timeout` does not cover it. #2132 tracks the same change across the other 31 write sites.

**Path building is shared.** `services/artworkSourcePath.ts` is used by both the service and the fixer, so a backfilled item and a derived one resolve identically.

## Test plan

- [x] `XmlTvWriter.test.ts` — derives a poster URL for a program with no artwork rows (plex and jellyfin); a stored row still wins over derivation; returns `undefined` for a `local` source and when `externalKey` or `mediaSourceId` is missing; derives from the show under `useShowPoster` and falls back to the episode when the show is not derivable; derives a track poster from its album. All 7 fail against the previous commit.
- [x] `BackfillProgramArtworkFixer.test.ts` — backfills programs with no artwork; idempotent across runs; resumes and picks up programs added later; pages past the 500-row batch boundary (501 programs); skips programs whose media source has an unusable URI without stalling the scan; **stops at the read budget even when nothing is written**; **spends one budget across both scans, not one each**.
- [x] `ArtworkService.test.ts` — derives a Plex URL with the token in the header and **not** in the URL; derives a Jellyfin URL with `X-Emby-Token`; returns `not-found` for no media source, for a `local` source, and for a missing program. The pre-existing `serveArtwork` guards are restored rather than replaced: never redirects a credentialed URL even with `proxyArtwork` disabled (GHSA-h3r4-r2f2-qf59), still redirects a credential-free one, keeps the access token in a header for plex/jellyfin/emby, attaches no credentials for a local source — plus a new case asserting a *derived* result is never redirected either.
- [x] Full server suite: 1572 passed, 3 skipped.
- [x] `pnpm lint-changed` clean.
- [x] Typecheck: build config is clean at 0 errors; including test files, 487 pre-existing errors before and after, zero net new.

Manual check worth doing before merge:

- [ ] On a library with no `artwork` rows, confirm `/api/xmltv.xml` emits `<icon>` for programs the backfill has not reached
- [ ] Confirm startup no longer stalls on a large library

## Not included

**No completion state.** Both `LEFT JOIN` scans still run on every boot, and rows that can never be backfilled — a media source with an unusable URI — stay in the predicate forever, so each boot re-reads them up to the budget. The read cap makes that cost bounded and small rather than a full table scan, but it does not eliminate it. #2128's suggestion 2 is a follow-up.

**`/api/programs/:id/thumb` is still removed.** Guides published before v1.3.11 reference it and will keep 404ing. Newly written XMLTV uses the working `/artwork/:type` URLs. Whether to add a compatibility alias is a separate call.


chrisbenincasa (commented):
**Requesting changes.** The backfill work is solid and the diagnosis in the description is accurate, but the load-bearing claim — that this restores XMLTV guide thumbnails — does not hold. Recommend landing the fixer half and redoing the artwork half against the actual call site.

I verified this by checking out the branch, running the new tests (10/10 pass), running the server typecheck (489 pre-existing errors, zero in changed files, matching your claim), and tracing the call paths.

---

## 🔴 P0 — The XMLTV regression is not fixed

`XmlTvWriter.resolveArtworkUrl` is the **only** producer of `<icon>`/`<image>` (`XmlTvWriter.ts:242`). It reads the in-memory `artwork` relation and still returns `undefined` when no row matches (`XmlTvWriter.ts:296-308`). It never calls `ArtworkService`:

```
$ grep -rn "ArtworkService" XmlTvWriter.ts xmltvApi.ts TvGuideService.ts
(none)
```

So `deriveArtworkFromSource` is unreachable for guide output. The guide builds rows via `TvGuideService.materializeGuideItem` → `programDB.getProgramsByIds` (artwork eager-loaded); a program with no artwork row gets `artwork: []` → `resolveArtworkUrl` returns `undefined` → **the `<icon>` is still dropped**.

The new fallback only reaches the HTTP route `/api/programs/:id/artwork/:type` — and that URL is only *emitted* once a stored row already exists. On the XMLTV path the new code never runs.

This branch's own test file pins the unchanged behaviour:

```ts
// XmlTvWriter.test.ts — on main AND on this branch
test('returns undefined when artwork array is empty', () => { ... });
test('returns undefined when artwork is undefined', () => { ... });
```

That is exactly what #2129 asked for, and what the PR description says was done ("`XmlTvWriter.resolveArtworkUrl` went from always returning a URL to returning `undefined`... The backfill becomes a way to persist and cache those paths"). The two halves are in different layers. The unticked manual check in the test plan — confirm `/api/xmltv.xml` emits `<icon>` — would have caught this.

**Also missed from #2129:** the `track` → album-uuid case. The track candidate only looks at `program.album?.artwork` (`XmlTvWriter.ts:283-289`); the old fallback redirected to the album uuid and nothing replaces it.

**Suggested shape:** put the derivation where the writer can use it. Either
(a) restore a fallback inside `resolveArtworkUrl` that synthesises `{{host}}/api/programs/:id/artwork/poster` whenever the program has a remote `sourceType` + `externalKey` — no DB lookup needed at emit time, since the endpoint already derives on request; or
(b) inject `ArtworkService` into `XmlTvWriter` and resolve per program.

(a) is cheaper and keeps the writer synchronous. Add a test asserting a program with `artwork: []` but `sourceType: 'plex'` + `externalKey` still yields a URL, and that a local source still yields `undefined`.

On `/api/programs/:id/thumb`: treating the alias as out of scope is defensible, but #2129 lists it as an explicit ask and every already-published XMLTV file contains those URLs. Worth a tracked follow-up rather than prose in "not included".

---

## 🟠 P1 — Four tests deleted, two of them security regression guards

`ArtworkService.test.ts` is a rewrite, not an extension: 111 lines removed, 102 added. Lost with no replacement:

| Test | What it guarded |
|---|---|
| `never redirects a credentialed URL, even with proxyArtwork disabled` | GHSA-h3r4-r2f2-qf59 token leak via `Location` |
| `still redirects a credential-free URL when proxyArtwork is disabled` | the non-credentialed redirect path |
| `keeps the %s access token out of the URL and in a request header` (3 cases) | token as header, never query param |
| `does not attach credentials for a local media source` | no credentials for local |

The header-vs-URL property is now covered incidentally (two assertions, and only for *derived* paths), but **nothing covers `serveArtwork` at all** anymore. That method is unchanged here and sits on the request path for every artwork fetch — including the ones this PR newly enables. Please restore the `serveArtwork` tests alongside the new derivation tests rather than instead of them:

```ts
test('never redirects a credentialed URL, even with proxyArtwork disabled', async () => {
  const redirect = vi.fn();
  const reply = { redirect, status: vi.fn().mockReturnThis(),
                  headers: vi.fn().mockReturnThis(), send: vi.fn().mockReturnThis() } as unknown as FastifyReply;
  await makeService({ program: undefined }).serveArtwork(
    { kind: 'url', url: 'http://plex.local/thumb', headers: { 'X-Plex-Token': 'tok' } }, reply);
  expect(redirect).not.toHaveBeenCalled();
});
```

---

## 🟠 P1 — The per-run cap bounds *writes*, not work; the loop can still spin unbounded

Both loops guard on `written` (`BackfillProgramArtworkFixer.ts:61`, `:109`), which only counts *successfully derived* rows:

```ts
written += this.insertArtwork(artworkRecords);   // 0 when every row is unresolvable
if (batch.length < SELECT_BATCH_SIZE) break;
await throttle();
```

A run that processes many programs and derives **zero** paths never trips `written < MAX_ROWS_PER_RUN`, never breaks, and walks the rest of the table in 500-row batches while inserting nothing. The new test actually constructs this state ("skips programs whose media source has an unusable uri") — it passes only because the row count is tiny. Those rows are permanently unbackfillable by construction: they stay in the `isNull(Artwork.uuid)` predicate forever, so **every boot** re-reads them all, forever. That is the #2128 symptom for this cohort, minus the inserts.

The comment at `:27-29` ("A backfill larger than this finishes over several restarts") states the intent, so the guard looks like a mismatch. Suggest counting reads separately:

```ts
let processed = 0;
while (processed < MAX_ROWS_PER_RUN) {
  ...
  processed += batch.length;
  written += this.insertArtwork(artworkRecords);
}
```

Related: the two loops each get their own 5k budget, so a boot can write 10k rows, not 5k.

Also consider `canRunInBackground = false` until the scan is genuinely bounded. The `runInternal` → `async` conversion is a real improvement (the first `await throttle()` after a full batch hands the loop back), but on a fresh install the first boot still does up to 500 synchronous reads + 5 × 100 synchronous inserts before that first yield — the worst case is the one users hit. The P1 in the description about the flag being misleading is only partially addressed.

---

## 🟡 P2 — No completion state (#2128 ask)

Still no completion marker, so both `LEFT JOIN` scans run on every boot in perpetuity, and the 5k cap means a 100k-program library needs ~20 restarts to converge — each paying the startup cost. #2128's suggestion 2 ("add completion state") is unimplemented. Being explicit that no cursor is persisted is honest, but the headline "bound the backfill" is really "rate-limit the backfill."

---

## 🟢 What's good

- **Diagnosis is correct and well-evidenced.** The `canRunInBackground` no-op analysis, the single-transaction write lock, and the cursor-free resume are all accurate.
- **`BEGIN IMMEDIATE`** is the right call and well-reasoned — `DEFERRED` read→write upgrade returning immediate `SQLITE_BUSY` outside `busy_timeout` is a real better-sqlite3 footgun.
- **`artworkSourcePath.ts`** is a genuine improvement; the shared-helper consistency argument holds.
- **Consistency check passed.** The derived path uses `Program.sourceType`/`externalKey` while the backfill uses `ProgramGroupingExternalId.externalKey`, but `ProgramGroupingMinter.ts:94,248,310` sets grouping `externalKey` from the same `externalId`, so both resolve identically.
- **`artworkToResult` uses `mediaSource.type`, not `entity.sourceType`**, for header injection — correct, and avoids a divergence bug.
- Tests pass (10/10), typecheck has zero errors in changed files, oxlint clean.

## Minor

- `deriveArtworkFromSource` calls `mediaSourceDB.getById`, then `artworkToResult` calls it again → two source lookups per artwork request. Pass the loaded source down.
- Derived results bypass `ImageCache` entirely (`cachePath: null`), so upstream is hit on every request. Acceptable, but cached and derived items now behave differently under load.

---

**Suggested split:** rebase the fixer changes (`BackfillProgramArtworkFixer.ts`, `artworkSourcePath.ts`, the new fixer test) as one PR — nearly ready once the cap counts reads — and move `ArtworkService.deriveArtworkFromSource` into a second PR together with the `resolveArtworkUrl` fallback that actually consumes it, restoring the `serveArtwork` tests in the process.

chrisbenincasa:
Thanks — the P0 is right, and it's the kind of miss that a passing test suite actively hides. Pushed a2571d25 addressing all three. Point by point below, including the two I did **not** do.

---

## 🔴 P0 — XMLTV regression — fixed

Took shape **(a)**. The derivation now lives in `resolveArtworkUrl`, so the writer stays synchronous and there's no DB lookup at emit time.

Rather than bolt a program-only fallback on the end, it re-walks the existing candidate list:

```ts
for (const candidate of candidates) {
  if (!candidate.id || !isDerivableArtworkSource(candidate.source)) continue;
  return `{{host}}/api/programs/${candidate.id}/artwork/poster`;
}
```

Each candidate carries a `source` ref alongside its `artwork`, so the stored-row pass and the derived pass see the same precedence — show poster under `useShowPoster`, then episode, then album for tracks, then the program. That falls out of reusing the list rather than being special-cased, and it's what restores the **`track` → album-uuid** case you flagged: the album candidate is already in the list, it just had nothing to match against.

`isDerivableArtworkSource` mirrors the guard in `deriveArtworkFromSource` — a source type `buildArtworkSourcePath` actually handles (`plex`/`jellyfin`/`emby`, not just `!== 'local'`, so a source type added later doesn't silently start emitting dead URLs), plus `externalKey` and `mediaSourceId`. When those are absent the endpoint would 404, so the `<icon>` is still dropped. The two "returns undefined when artwork is empty/undefined" tests you quoted still pass unchanged, because `makeProgram` defaults `mediaSourceId: null` — they now assert "not derivable → undefined" rather than "never derives", which is the behaviour I'd want pinned either way.

7 new tests in `XmlTvWriter.test.ts` covering plex, jellyfin, stored-wins-over-derived, local, missing `externalKey`/`mediaSourceId`, show-under-`useShowPoster`, show-not-derivable→episode, and track→album. I stashed the source changes and re-ran to confirm all 7 fail against d3165de3 — they aren't passing by accident.

## 🟠 P1 — deleted tests — restored

All four are back, alongside the derivation tests rather than instead of them, via a separate `makeStoredArtworkService` helper so the two styles don't fight. Added one more: **`never redirects a URL derived from a credentialed source`**. That path didn't exist when the original guard was written, and a derived result is credentialed too — it's exactly the surface this PR newly opens, so it should be pinned.

## 🟠 P1 — the cap bounds writes, not work — fixed

Counts reads now, and the budget is shared:

```ts
const budget = { remaining: this.maxRowsReadPerRun };
await this.backfillProgramArtwork(budget);
await this.backfillGroupingArtwork(budget);
```

with `.limit(Math.min(SELECT_BATCH_SIZE, budget.remaining))` so it can't overshoot, and `budget.remaining -= batch.length` right after the cursor advances. That also closes the 5k-each → 10k-per-boot gap.

One nuance on the framing: the old loop terminated rather than spinning — the uuid cursor advanced regardless of whether anything was written — so the failure was an unbounded *full table walk* per boot, not a hang. Same fix, and the consequence you describe for the unbackfillable cohort was real.

Testing it needed the constant to be reachable, so it's a `protected readonly maxRowsReadPerRun` defaulting to the module const, and the tests subclass with 50. Two new cases:

- `stops at the read budget even when nothing is written` — 120 unresolvable programs plus one backfillable program pinned at `ffffffff-…` so it sorts last. Only a scan that reads past the budget reaches it. Expects 0 rows written; fails against d3165de3, which reads to the end and writes 1.
- `spends one budget across both scans, not one each`.

## 🟡 P2 — completion state — not done

Agreed, and you're right that "bound the backfill" was overselling it. Rewrote the description: this is now explicitly under **Not included**, with the residual cost stated plainly — unbackfillable rows stay in the predicate and are re-read every boot, the read cap just makes that bounded and small instead of a full table scan. Worth its own PR against #2128 rather than growing this one.

## `canRunInBackground` — left as `true`, flagging it

I did not flip it. Your worst-case reasoning holds: a fresh install still does up to 500 synchronous reads plus 5 × 100 synchronous inserts before the first `await throttle()`. What changed is that it's now *one* batch of that rather than the whole table, and with the derivation fallback in place the fixer is no longer load-bearing for correctness — so the argument for letting it block startup is weaker, not stronger. Happy to flip it or drop `SELECT_BATCH_SIZE` to shorten the first yield if you'd rather; say which and I'll push it.

## Minor — both left alone deliberately

- **Double `mediaSourceDB.getById`** — real, and a one-liner to thread the loaded source through. Left it out to keep this diff scoped to the three findings; call it and I'll add it here or in the follow-up.
- **Derived results bypass `ImageCache`** — left as-is. Caching a derived path means writing the artwork row, which is the backfill's job; doing it opportunistically on the request path is a different design decision than "don't 404", and I'd rather not smuggle it in here.

---

Verification: 51 tests across the three files, full server suite 1572 passed / 3 skipped, `pnpm lint-changed` clean, build typecheck 0 errors, 487 pre-existing errors including test files both before and after.

Still unticked and not something I can exercise: `/api/xmltv.xml` emitting `<icon>` for un-backfilled programs against a real library, and startup on a large one.

chrisbenincasa (commented):
Re-reviewed at a2571d253. **All three blocking findings are addressed, and I verified each independently.** The two halves are now in the right layers and the budget actually bounds work.

## ✅ P0 resolved — guide artwork, at the right layer

`XmlTvWriter.resolveArtworkUrl` now emits a derived URL for a candidate with no stored row (`XmlTvWriter.ts:346-354`), after the stored-artwork loop. This is the fix I asked for: the writer decides *whether* to emit, the endpoint decides *what URL* to fetch. Built as a `{{host}}`-templated endpoint URL it stays synchronous — no DB lookup added to guide generation.

The `track` → album case from #2129 is covered too, via `source: program.album`.

I confirmed the two guards agree, since that was the risk with splitting the logic:

| | writer | endpoint (`deriveArtworkFromSource`) |
|---|---|---|
| sourceType | `['plex','jellyfin','emby']` | truthy → `buildArtworkSourcePath` default → `undefined` |
| externalKey | `isNonEmptyString` | truthy |
| mediaSourceId | `isNonEmptyString` | truthy |

Writer is the stricter subset, so everything it emits the endpoint can answer. The `useShowPoster` / episode / album precedence also matches (`show` → episode → album → program).

## ✅ P1 resolved — security tests restored

All four removed cases are back, plus a new one I'd have asked for anyway:

- `never redirects a credentialed URL, even with proxyArtwork disabled`
- `still redirects a credential-free URL when proxyArtwork is disabled`
- `keeps the %s access token out of the URL and in a request header` (plex/jellyfin/emby)
- `does not attach credentials for a local media source`
- **new:** `never redirects a URL derived from a credentialed source` — covers the path this PR newly enables, which is exactly the gap I flagged.

## ✅ P1 resolved — the cap counts reads

`MAX_ROWS_READ_PER_RUN` is now a single `ReadBudget` shared by both scans (`:69-79`), decremented by `batch.length` (`:106`, `:161`), with `.limit(Math.min(SELECT_BATCH_SIZE, budget.remaining))` so a partial final batch can't overshoot. That closes the "writes stay at zero, budget never trips" hole.

The two tests pin precisely the property I asked for, and they're well-constructed:

- `stops at the read budget even when nothing is written` — 120 unusable programs plus a backfillable row at `ffffffff-...`, which sorts last. Reaching it would insert, so `artworkCount === 0` proves the scan stopped rather than exhausted the table. Nice use of the sort order to make the assertion meaningful.
- `spends one budget across both scans, not one each` — 80 programs with `BUDGET = 50` yields 50, not 80. This catches the per-scan-budget regression too.

The `BoundedFixer` subclass override is the right call over inserting 5k rows.

## Verification

- **51 tests pass** across the three touched suites (`XmlTvWriter.test.ts`, `ArtworkService.test.ts`, `BackfillProgramArtworkFixer.test.ts`), up from 10.
- **Typecheck: identical to `main`.** 489 errors on both branches, and `XmlTvWriter.test.ts` has exactly 13 on both — no net new errors. (Those 13 are pre-existing fixture-type friction inherited from `createFakeProgram`/`makeProgram`; the new tests don't add to them.)
- No `as any`, no new unguarded `!` in changed files.
- `canRunInBackground` is now honest: the first `await throttle()` after a full batch genuinely returns to the loop.

## Two non-blocking notes

1. **`DERIVABLE_SOURCE_TYPES` in `XmlTvWriter.ts:44` duplicates the switch in `artworkSourcePath.ts`.** Today they're a strict subset so they can't disagree, but a future source type appended to `buildArtworkSourcePath` would derive at the endpoint while the writer silently keeps dropping `<icon>` — a quiet regression in the direction that matters. Exporting the supported set from `artworkSourcePath.ts` and importing it in both places would make that impossible.

2. **No test drives the emitted URL through `resolveArtwork` end to end.** The writer test asserts the URL string and `ArtworkService.test.ts` asserts derivation separately, with `{{host}}` left in place. The `{{host}}` → real-host substitution lives in the serving path (as with the pre-existing stored-artwork URLs), so this is consistent with how the feature already worked — but a single integration test fetching an emitted URL would lock the contract between the two halves, which is the seam that broke in #1923 in the first place.

## On the #2128 asks not taken

Not persisting a cursor and not adding `FixerRunner` completion state are fair scope calls, and the changed comment now says plainly that a partial backfill isn't user-visible — which is true *because* the writer fallback landed. Worth filing the completion-state follow-up separately so the two `LEFT JOIN` scans eventually stop running on every boot.

LGTM once you're happy with the two notes. Nice turnaround — the writer-side derivation is a smaller change than I expected and it lands in the right place.

PR #2143, opened 2026-09-22
fix(xmltv): emit program elements in DTD order

h0lylag:
## Problem

I ran into this while setting up Tunarr with Jellyfin 12.1 Live TV. The channels and guide showed up, but every show had a placeholder image. Jellyfin could reach the artwork URLs, and updating, re-adding shows, and refreshing the guide didn't help.

The artwork was in the feed, but Tunarr was putting `<image>` right before `<icon>`, which is out of order according to the [XMLTV DTD](https://github.com/XMLTV/xmltv/blob/b424ef772a667856ca44b9dac3b4310d02191dea/xmltv.dtd#L301-L306). Jellyfin also has a parser bug: [`ProcessImageNode` already moves past the image](https://github.com/jellyfin/Jellyfin.XmlTv/blob/6c8c302c8520e5f41cd51dd392ac631f97179d82/src/Jellyfin.XmlTv/XmlTvReader.cs#L981-L1025), but [the caller runs `Skip()` afterward](https://github.com/jellyfin/Jellyfin.XmlTv/blob/6c8c302c8520e5f41cd51dd392ac631f97179d82/src/Jellyfin.XmlTv/XmlTvReader.cs#L291-L295). With Tunarr's compact XML, that skips the icon Jellyfin needs for the thumbnail.

I got the thumbnails working by patching Jellyfin locally to remove the extra `Skip()` call. The patched parser read all 749 program icons in the captured guide, while the original parser read none. This PR fixes the element order in Tunarr so other users can resolve the same artwork issue without patching Jellyfin.

## Fix

`@iptv/xmltv` writes child elements in the order their properties are added. Swapping `<icon>` and `<image>` would restore the thumbnails in unpatched Jellyfin, but it would leave other elements out of XMLTV order. To fix the artwork issue and put each program's child elements in DTD order, this reorders the existing assignments in `XmlTvWriter`: length after categories and keywords, icons before episode numbers, ratings after episode and video metadata, and images last.

The artwork lookup moves earlier so its URL is available when the icon is added. Both artwork elements still use that same URL, and the metadata values stay the same.

Jellyfin still needs its own fix for the extra skip, since that can drop elements from correctly ordered feeds too.

## Test plan

- [x] Added eight cases in `XmlTvWriter.test.ts` that check the XML emitted by `@iptv/xmltv`'s actual serializer: episodes, movies, and tracks with and without artwork, plus cases with missing optional metadata. They cover element order, escaping, duration units, and artwork preservation. All 49 XMLTV tests passed.
- [x] Ran the same tests against the original writer from `main`. Seven new cases failed on element order, and the other 42 passed. All 49 passed with this fix.
- [x] Tested feeds generated from the same 24-program fixture in two fresh, unpatched Jellyfin 12.1.0 containers, using the original writer for one and this branch for the other. Jellyfin picked up **0/12 posters** with the original feed and **12/12** with the fixed feed. All 12 posters were served successfully through Jellyfin's image endpoint. Both instances imported all 24 programs, and the 12 entries without artwork stayed empty. Both containers used the same unmodified parser revision, `6c8c302`.
- [x] Passed `pnpm turbo typecheck`, `pnpm turbo build --filter=@tunarr/web`, and `pnpm lint-changed --branch`. Formatting and commit message checks passed too.
- [ ] Full suite (`pnpm turbo test -- run`): 1,854 passed, 2 skipped, 2 failed.

The two failures are in existing `ArtworkService.test.ts` cases at lines 223 and 258; this PR doesn't change them. Both make unmocked requests to `plex.local` and hit the 5-second timeout in the full run and a focused rerun. The rerun also logged `ENOTFOUND plex.local`. The first failure was also [reported in #2112 and reproduced twice on unmodified `dev`](https://github.com/chrisbenincasa/tunarr/pull/2112#issuecomment-5720781954).


chrisbenincasa:
Sorry for the delay here, this got lost in the mix.

chrisbenincasa (approved):

h0lylag:
Thank you for the quick review and merge. :beers:

